PR #5, opened 2017-11-13
Build on Linux and making UI not bright blue

cmiceli:
This includes changes which we required in order to facilitate building on Arch Linux, tidying up the UI and making things look a bit more presentable. There is another change to the backend which is what I needed to do to run the app (was causing fatals).

Sorry for > 1 change in the request, not worth pulling apart.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -9,7 +9,8 @@
 	"dependencies": {
 		"apollo-client": "^1.9.0",
 		"buffer": "^5.0.7",
-		"graphql-tag": "^2.4.2",
+		"expo": "^22.0.3",
+		"graphql-tag": "^2.5.0",
 		"immutability-helper": "^2.3.0",
 		"lodash": "^4.17.4",
 		"moment": "^2.18.1",
--- a/client/src/config.js
+++ b/client/src/config.js
@@ -1,3 +1,3 @@
 
 
-export const GRAPHQL_ENDPOINT = 'http://10.1.0.59:8080/graphql';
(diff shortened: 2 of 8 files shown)


client/src/config.js
```diff
@@ -1,3 +1,3 @@
 
 
-export const GRAPHQL_ENDPOINT = 'https://ses-availability-api.herokuapp.com/graphql';
+export const GRAPHQL_ENDPOINT = 'http://10.1.0.59:8080/graphql';
```
sdunster: revert

client/package.json
```diff
@@ -9,7 +9,8 @@
 	"dependencies": {
 		"apollo-client": "^1.9.0",
 		"buffer": "^5.0.7",
-		"graphql-tag": "^2.4.2",
+		"expo": "^22.0.3",
+		"graphql-tag": "^2.5.0",
```
sdunster: whats this?

cmiceli: needed to build this on linux earlier. can probably fuck it off.

client/src/config.js
cmiceli: fair enough

cmiceli:
screenshot_20171112-212445

cmiceli:
not the best screenshot, but with some content it will be molto bene

OSPFNeighbour:
You guys are hilarious. Chris - Sam will show you UI mock ups as reference
for nav bar. We need to drop a few icons and consolidate some pages.

PR #9, opened 2017-11-14
Added tags and capabilities to model.

OSPFNeighbour:
Added tags and capabilities, expanded several other models to match up with Data Model POC

```json
{
  "data": {
    "user": {
      "organisation": {
        "name": "NSW SES"
      },
      "username": "chris",
      "email": "test@miceli.net.au",
      "devices": [
        {
          "uuid": "1234abc"
        }
      ],
      "tags": [
        {
          "name": "Geographical Area 1"
        }
      ],
      "capabilities": [
        {
          "name": "Capability 1"
        }
      ],
      "groups": [
        {
          "name": "Bankstown",
          "tags": [
            {
              "name": "Geographical Area 1"
            }
          ],
          "schedules": [
            {
              "name": "Bankstown Roster",
              "details": "Weekly roster",
              "deeplink": "https://foobar.com/"
            },
            {
              "name": "Bankstown Primary School Fete",
              "details": "5 people needed for school fete",
              "deeplink": "https://bankstownSES.com/forums/341231"
            }
          ],
          "events": [
            {
              "name": "Real Time Event #123",
              "details": "thing with the stuff",
              "deeplink": "https://foobar/event/123"
            }
          ]
        }
      ]
    }
  }
}
```

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -23,7 +23,7 @@ function getAuthenticatedDevice(ctx) {
 }
 
 export const locationHandler = {
-  updateLocation(_, args, ctx){
+  updateLocation(_, args, ctx) {
     return getAuthenticatedDevice(ctx).then((device) => {
       const { locationLat, locationLon } = args.location;
       return device.update({
@@ -54,7 +54,7 @@ export const deviceHandler = {
 }
 
 export const userHandler = {
-  query(_, args, ctx){
+  query(_, args, ctx) {
     return getAuthenticatedUser(ctx);
   },
   createUserX(_, args, ctx) {
@@ -93,24 +93,24 @@ export const userHandler = {
      }));
    });
   },
-  devices(user, args, ctx){
+  devices(user, args, ctx) {
     return Device.findAll({
       where: { userId: user.id, }
     })
   },
-  tags(user, args, ctx){
+  tags(user, args, ctx) {
      return user.getTags();
   },
-  capabilities(user, args, ctx){
+  capabilities(user, args, ctx) {
      return user.getCapabilities();
   },
 }
 
 export const scheduleHandler = {
-  timeSegments(schedule, args, ctx){
+  timeSegments(schedule, args, ctx) {
     return user.getTimeSegments();
   },
-  createSchedule(_, args, ctx){
+  createSchedule(_, args, ctx) {
     return Schedule.create({ name: args.schedule.name }).then((schedule) => {
       return Group.findById(args.schedule.group_id).then((group) => {
         schedule.setGroup(group);
@@ -131,6 +131,12 @@ export const organisationHandler = {
   users(org, args, ctx) {
     // TODO: think about who we show the complete organisation user list to
     return User.findAll({ organisationId: org.id })
+  },
+  tags(org, args, ctx) {
+    return org.getTags();
+  },
+  capabilities(org, args, ctx) {
+    return org.getCapabilities();
   }
 }
 
@@ -156,11 +162,11 @@ export const groupHandler = {
     console.log(args.groupUpdate);
     return getAuthenticatedUser(ctx).then(() => {
       return Group.findById(args.groupUpdate.group_id).then((group) => {
-        if (!group){
+        if (!group) {
             return Promise.reject("No group!");
         }
         User.findById(args.groupUpdate.user_id).then((user) => {
-            if (!user){
+            if (!user) {
                 return Promise.reject("No user!");
             }
             group.addUser(user).then(() => {return group;});
@@ -168,7 +174,7 @@ export const groupHandler = {
       })
     })
   },
-  tags(group, args, ctx){
+  tags(group, args, ctx) {
      return group.getTags();
   },
 }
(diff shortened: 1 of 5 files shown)


server/src/models.js
```diff
@@ -64,6 +76,14 @@ UserModel.belongsTo(OrganisationModel);
 // devices belong to a single user
 DeviceModel.belongsTo(UserModel);
 
+//users -> tags
+UserModel.belongsToMany(TagsModel, { through: 'tag_user' });
```
sdunster: I think you want to swap these around:
```
TagsModel.belongsToMany(UserModel,{ through: 'user_tag' });
```
ultimately it might work out the same in the DB but I think it makes it easier to reason about. (the _user_ has the tag, not the _tag_ has the user)

server/src/models.js
```diff
@@ -64,6 +76,14 @@ UserModel.belongsTo(OrganisationModel);
...
+
+//groups -> tags
+GroupModel.belongsToMany(TagsModel, { through: 'tag_group' });
```
sdunster: as above

server/src/models.js
```diff
@@ -64,6 +76,14 @@ UserModel.belongsTo(OrganisationModel);
...
+
+//users belong to capabilities tags
+UserModel.belongsToMany(CapabilitiesModel, { through: 'capabilities_user' });
```
sdunster: as above

server/src/models.js
```diff
@@ -103,25 +125,52 @@ db.sync({force: true}).then(() => {
...
           }),
-
           // create a device
           Device.create({
             uuid: '1234abc',
           }).then((device) => {
             device.setUser(user);
           }),
-
+          // create a capabilitie
```
sdunster: capability

server/src/models.js
```diff
@@ -78,6 +98,8 @@ TimeSegmentModel.belongsTo(UserModel);
 const Organisation = db.models.organisation;
 const Group = db.models.group;
 const User = db.models.user;
+const Capabilities = db.models.capabilities;
```
sdunster: this should use the singular 'Capability'

server/src/models.js
```diff
@@ -78,6 +98,8 @@ TimeSegmentModel.belongsTo(UserModel);
...
+const Tags = db.models.tags;
```
sdunster: as above 'Tag'

server/src/models.js
```diff
@@ -27,6 +27,14 @@ const UserModel = db.define('user', {
   version: { type: Sequelize.INTEGER }, // version the password
 });
 
+const CapabilitiesModel = db.define('capabilities', {
```
sdunster: use singular `CapabilityModel` and `'capability'` for the table name

server/src/models.js
```diff
@@ -27,6 +27,14 @@ const UserModel = db.define('user', {
...
+  name: { type: Sequelize.STRING },
+});
+
+const TagsModel = db.define('tags', {
```
sdunster: as above

sdunster:
Oh one more thing... The tags are global to the system. Like everything else they really should be per-organisation so that organisation A would not see the tags that organisation B has defined.

cmiceli:
So something which I think is worth considering from a code perspective, we might want to generalise the concept of a tag.

have a table of tags, and a table of associations (I'll call it TagAssociation) which have the concept of a scope (due to the non-unique id for tables). So the proposal is something like:

TagAssociation:
  - tag_id
  - scope (probably an enum which represents groups, users, etc)
  - object_id (id for the tagged object)

so with the scope and object_id we can now arbitrarily look everything up.

I think ultimately, a good solution would involve handing out unique id's across all tables but it might be a bit of over-engineering.

OSPFNeighbour:
Ill address Sam's thing's first, for my own sake of self learning. Then delete everything and have a play on Chris' idea.

sdunster:
I actually disagree with Chris's idea. I think it's much easier to just define seperate relationship tables to each of the things we want to attach tags to. One central tag table definitely makes sense though.
TL;DR I don't think you should change what you have now other than maybe unifying `Capability` with `Tag` although I don't think even that is worthwhile.

OSPFNeighbour:
@sdunster regarding making them under the org, no idea how do that. flat DB i would store the id of the org against the tag... but dunno how do carry that concept over.

OSPFNeighbour:
@sdunster deeplink is your term! you defined it on the data model.... a link back to the source of the request, for an event in the ses it would be a beacon link to the job.

sdunster:
Re deeplink I assumed it was meant for locating the event or something. Lets just leave that stuff out and add it when we have a use for it.
Re making tags belong to an org - it's literally exactly that.
```
TagModel.belongsTo(OrganisationModel)
```

server/src/logic.js
```diff
@@ -1,4 +1,4 @@
-import { Organisation, Group, User, Device, Event, Schedule, TimeSegment } from './models' ;
+import { Organisation, Group, User, Device, Event, Schedule, TimeSegment, Tags, Capabilities } from './models' ;
```
sdunster: you didn't test this did you? `Capability` 👎

OSPFNeighbour:
land this when ready if you are happy to expand to this.

PR #11, opened 2017-11-14
organisation resolver

OSPFNeighbour:
I might be totally wrong here, but I think ive added a resolver for organisation -> groups

```
{
  "data": {
    "user": {
      "organisation": {
        "groups": [
          {
            "name": "Bankstown"
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -118,6 +118,10 @@ export const organisationHandler = {
   createOrganisation(_, args, ctx) {
     return Organisation.create(args.organisation);
   },
+  groups(organisation, args, ctx) {
+    console.log(args)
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
any good?

server/src/logic.js
```diff
@@ -118,6 +118,10 @@ export const organisationHandler = {
   createOrganisation(_, args, ctx) {
     return Organisation.create(args.organisation);
   },
+  groups(organisation, args, ctx) {
+    console.log(args)
+    return Group.findAll({ groupId: organisation.id });
```
sdunster: for reference this was good except:
```
return Group.findAll({ orgnisationId: organisation.id });
```

PR #16, opened 2017-11-14
Placeholder Home Screen

ajshort:
I also extracted some repeated styles into a separate file.

### DIFF
--- a/client/src/screens/home.screen.js
+++ b/client/src/screens/home.screen.js
@@ -1,22 +1,10 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import {
-  FlatList,
-  ActivityIndicator,
-  Button,
-  Image,
-  StyleSheet,
-  Text,
-  TouchableHighlight,
-  View,
-} from 'react-native';
-import { graphql, compose } from 'react-apollo';
-import moment from 'moment';
+import { StyleSheet, Text, View } from 'react-native';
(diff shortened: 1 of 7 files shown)


client/src/screens/home.screen.js
```diff
@@ -1,108 +1,160 @@
...
+  eventUrgent: {
+    borderColor: 'red'
+  }
 });
 
 class Home extends Component {
   static navigationOptions = {
     title: 'Home',
-    tabBarIcon: ({ tintColor}) => <Icon size={30} name={'home'} color={tintColor} />
+    tabBarIcon: ({ tint }) => <Icon size={48} name='home' color={tint} />
```
sdunster: reverting this size increase cause it looks dumb on iOS. Clearly we just need to handle the styles differently on iOS vs android but I have more important things to work on right now. Feel free to take it on if you'd like.

sdunster:
Landed in b0723b1

client/src/screens/home.screen.js
ajshort: Good spot, the size change wasn't deliberate - I was probably trying to fiddle with the power icon size.

sdunster: The variable name change here actually broke the tint colour being applied as well. This is a JS unpacking construct so the name needs to match what is in the argument passed to this func. There's ways to get around this but you can google.

PR #17, opened 2017-11-15
Add has-many relation from user to devices

ajshort:
Was getting some errors with the server and devices not being associated to user.


OSPFNeighbour:
Two fixes for the same issue, Both would work but at this stage im unsure if we want to change the data relationship.

Duplicate fix as #20
Addresses #19

ajshort:
I don't think this changes the relationship? It just makes sure both ends are properly defined.

PR #18, opened 2017-11-15
allow express to bind to specific address.

OSPFNeighbour:
yea... that.

### DIFF
--- a/server/src/main.js
+++ b/server/src/main.js
@@ -11,6 +11,7 @@ import { User, Device } from './models';
 import { executableSchema } from './schema';
 
 const GRAPHQL_PORT = 8080;
+const GRAPHQL_ADDRESS = '0.0.0.0'
 const GRAPHQL_PATH = '/graphql';
 
 const port = process.env.PORT ? process.env.PORT : GRAPHQL_PORT;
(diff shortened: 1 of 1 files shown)


server/src/main.js
```diff
@@ -36,6 +37,6 @@ app.use('/graphiql', graphiqlExpress({
 
 const graphQLServer = createServer(app);
 
-graphQLServer.listen(port, () => {
+graphQLServer.listen(port,GRAPHQL_ADDRESS, () => {
```
sdunster: plz 2 whitespace
```
graphQLServer.listen(port, GRAPHQL_ADDRESS, () => {
```

OSPFNeighbour:
Windows seems to bind differently, change was made due to someone reporting
issues with default bind.

0.0.0.0 was a holder as I listed a specific address I wanted to bind to.

sdunster:
Please figure out the details so we can make sure this is actually the right thing to do. It might be better just to special case this for windows if absolutely necessary.

sdunster:
Luke Quinane - Yesterday at 1:47 PM
> re the windows binding issue. I've double checked that and it was me just getting mixed up between the graphql server which was bound to 8080 and something else which was bound to 8081

I don't think we need this change so I'm going to close it but please feel free to reopen if you think it's important.

PR #20, opened 2017-11-15
Fix signup devic bind error

OSPFNeighbour:
relationship between user and device is

DeviceModel.belongsTo(UserModel);

This query was written for the reverse logic.


OSPFNeighbour:
any reason you are holding off on landing?

OSPFNeighbour:
Rolled into #31

Issue #19, opened 2017-11-15
Device is not associated to user on user signup and errors are not returned to user


OSPFNeighbour:
Unhandled rejection SequelizeEagerLoadingError: device is not associated to user!

when a user signs up

sdunster:
We need to make sure the following return all promises back to be resolved by graphql so that all errors are caught and handled appropriately and it actually fails the tests instead of silently succeeding:
* login mutation
  * load user
  * device creation/check
  * jwt generation (not sure if async or not)
* signup mutation
  * user creation
  * device creation
  * jwt generation (not sure if async or not)

https://github.com/NSWSESMembers/availability-poc/blob/master/server/src/resolvers.js#L52

sdunster:
Pretty sure this is no longer a thing - reopen if it is.

PR #21, opened 2017-11-15
Add a new test for creating a user.

tmyroadctfig:
I was trying to add a test to capture the problem in #19, but it seems to pass either way. Not sure if that means that the error is not bubbling up properly.

### DIFF
--- a/server/tests/index.test.js
+++ b/server/tests/index.test.js
@@ -26,7 +26,7 @@ describe('GraphQL Queries', function() {
     done();
   });
 });
-  it('should return test devic uuid', done => {
+  it('should return test device uuid', done => {
   self
   .test(
     JSON.stringify({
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
you can add a catch to the end of the promise but im still not sure how we tell Graphql to fail the query.

Normally i spend half my time making this fails in a nice way, this is the opposite for once.

OSPFNeighbour:
for your test you could ask graphql to return the id or something, assuming that would be missing when the query fails.

sdunster:
Pretty sure what you actually want to do is return a promise to mocha and have it resolve it. You don't need to add an error case unless you want to handle it. If you let the error fall through then the test will fail if there is an error. I will add an example inline.

server/tests/index.test.js
```diff
@@ -94,4 +94,36 @@ describe('GraphQL Queries', function() {
     done();
   });
 });
+  it('should create a new user', done => {
```
sdunster:
```
it('should create a new user', () => {
  // we RETURN the resulting promise for Mocha to resolve
  return self
    .test(
      JSON.stringify({
        ...
      }),
    )
    .then(res => {
      expect(res.status).toBe(200);
      expect(res.success).toBe(true);
      // you don't need to call done()
    });
    // no error case - mocha will complain if it gets one
});
```

sdunster:
If you wanted to test the error case specifically just raise an error in your `then()` and write a normal `catch()` that verifies the error is what you expect. Because you have implemented `catch()` the error will never reach mocha so the test will pass.

btw this follows what we do in libbeacon tests. You can read more about it here: http://tobyho.com/2015/12/16/mocha-with-promises/

OSPFNeighbour:
Err, pretty sure GraphQL isn't raising any error. As in sequaliser is
failing on a query but GraphQL just lets it and returns no error status or
message.

Same from the app, it gets no error even though half its query failed.

sdunster:
OK so to cover this situation we need to write a test that creates a user and then runs a query to confirm the device is registered to that user.

OSPFNeighbour:
Duplicate to #31

sdunster:
Please close if no longer relevant.

PR #23, opened 2017-11-15
sets org to default id=1

### LINKED ISSUE #19: Device is not associated to user on user signup and errors are not returned to user

OSPFNeighbour:
Fixes #19

### DIFF
--- a/server/src/resolvers.js
+++ b/server/src/resolvers.js
@@ -44,7 +44,6 @@ export const Resolvers = {
         if (!existing) {
           // hash password and create user
           return bcrypt.hash(password, 10).then(hash => User.create({
-
             email,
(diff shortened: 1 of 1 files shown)


server/src/resolvers.js
```diff
@@ -49,6 +49,7 @@ export const Resolvers = {
             username: username,
             version: 1,
           })).then((user) => {
+            user.setOrganisation(1); //default org for now
```
sdunster: pretty sure this returns a promise so we need to return that so it can be awaited by graphql. I'll rewrite this whole section in #19.

PR #31, opened 2017-11-17
Tdykes.usermgmt

OSPFNeighbour:
Partial fix for #19.. might be a full fix.. dunno

Support promises for

=> signup mutation

Added mutation for

=> deleteUser

Added/Fixed tests for:

creating test user
deleting test user

### DIFF
--- a/server/src/resolvers.js
+++ b/server/src/resolvers.js
@@ -44,7 +44,6 @@ export const Resolvers = {
         if (!existing) {
           // hash password and create user
           return bcrypt.hash(password, 10).then(hash => User.create({
-
             email,
             password: hash,
             username: username,
--- a/server/src/resolvers.js
+++ b/server/src/resolvers.js
@@ -44,11 +44,13 @@ export const Resolvers = {
         if (!existing) {
           // hash password and create user
           return bcrypt.hash(password, 10).then(hash => User.create({
+
             email,
             password: hash,
             username: username,
             version: 1,
           })).then((user) => {
+            user.setOrganisation(1); //default org for now
             deviceHandler.addDevice(user, deviceId);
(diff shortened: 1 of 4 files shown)


server/src/resolvers.js
```diff
@@ -34,6 +34,19 @@ export const Resolvers = {
       console.log(args);
       return userHandler.createUserX(_, args, ctx);
     },
+    deleteUser(_, DeleteUserInput, ctx) {
```
sdunster: I know you're trying to flesh the API out but I don't see a use for this. I see 2 situations where we would delete a user:
1. admin user needs to delete a user - input would be an ID number and honestly is probably just going to do it via SQL.
2. User wants to delete themselves. No ID number needed because their auth token identifies them.

server/src/resolvers.js
```diff
@@ -49,12 +62,18 @@ export const Resolvers = {
...
-            return user;
+            return user.setOrganisation(1).then(() => {
+              return deviceHandler.addDevice(user, deviceId).then(() => {
+                const { id } = user;
+                const token = jwt.sign({ id, device: deviceId, email, version: 1 }, JWT_SECRET);
+                user.authToken = token;
+                ctx.user = Promise.resolve(user);
+                return user;
+              });
+            }).catch(function(err) {
```
sdunster: }).catch((err) => {

server/src/resolvers.js
OSPFNeighbour: Testing.

I cant test that the code to create and catch duplicate users works if I cant delete the user programatically. I was having to flush the DB by hand every time i ran a of the signup mutation.

sdunster:
See #36

PR #36, opened 2017-11-21
Massive code cleanup + #31

sdunster:
@OSPFNeighbour adds:
* signup mutation
* deleteUser mutation
* creating test user
* deleting test user

@sdunster adds:
* Dont allow sign up with duplicate email or username
* Move signup and login to logic.js
* Establish 2 way relationships between all models
* Rewrite signup/login with sane promise handling
* Tests that use promises
* Extensive testing of all GraphQL query resolvers
* De-lint most of the server code
* Move all model creation code to `Creators` - implement this in all logic and
  test-data creation.
* Wrote a bunch of validators to make sure all entities are created with
  all of the appropriate associations.
* Create `test-data` which uses `Creators` and is responsible for creating
  test data on boot.
* Found a bunch of places we don't return `Promises` properly - fixed.
* Implement `TimeSegments` although we don't have any created automatically
  yet.
* Add association between Groups and Organisations - was previously missing.
* Renamed some graphQL fields to use `camelCase` instead of `under_scores`.
* Device-generated ID fields are always referred to as `deviceUuid` now to
  avoid conflict with `deviceId`.


OSPFNeighbour:
Looks ok at quick glance.

Issue #22, opened 2017-11-15
Can't pull to refresh if there's no items in the list


sdunster:
This affects (tabs):
* groups
* schedules
* events

Root cause:
we don't display an empty list as a list - it is just a plain view. Lets make the list smarter and have it show a different cell when the list is empty so that we can still support pull to refresh.

OSPFNeighbour:
Fixed though smarter UI items

PR #37, opened 2017-11-21
Tdykes.ui.wip

OSPFNeighbour:
Big UI changes to support groups.

Minor changes to backend to support groups

### DIFF
--- a/client/src/graphql/all-groups.query.js
+++ b/client/src/graphql/all-groups.query.js
@@ -1,0 +1,24 @@
+import gql from 'graphql-tag';
+
+// get the user and all user's groups
+export default gql `
+  query {
+    user {
+      id
+      organisation {
+         groups {
+            name
+            id
+            tags {
+              name
+              id
+           }
+         }
+    }
+    groups {
+      id
+      name
+    }
+  }
+}
+`;
--- a/client/src/graphql/all-groups.query.js
+++ b/client/src/graphql/all-groups.query.js
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 
 // get the user and all user's groups
-export default gql `
+export default gql`
   query {
     user {
       id
@@ -14,11 +14,11 @@ export default gql `
               id
            }
          }
-    }
-    groups {
-      id
-      name
+      }
+      groups {
+        id
+        name
+      }
     }
   }
-}
 `;
--- a/client/src/graphql/join-group.mutation.js
+++ b/client/src/graphql/join-group.mutation.js
@@ -1,0 +1,9 @@
+import gql from 'graphql-tag';
+
+export default gql`
+  mutation addUserToGroup($groupUpdate: AddUserToGroupInput!) {
+    addUserToGroup(groupUpdate: $groupUpdate) {
+      id
+    }
+  }
+`;
--- a/client/src/navigation.js
+++ b/client/src/navigation.js
@@ -7,11 +7,14 @@ import { REHYDRATE } from 'redux-persist/constants';
 
 import Home from './screens/home.screen';
 import Groups from './screens/groups.screen';
+import Group from './screens/group.screen';
 import Events from './screens/events.screen';
 import Schedules from './screens/schedules.screen';
 import Signin from './screens/signin.screen';
 import Settings from './screens/settings.screen';
 import NewGroup from './screens/new-group.screen';
+import SearchGroup from './screens/search-groups.screen';
+
 
 const tabBarConfiguration = {
   tabBarPosition: 'bottom',
@@ -41,7 +44,10 @@ const MainScreenNavigator = TabNavigator({
 const AppNavigator = StackNavigator({
   Main: { screen: MainScreenNavigator },
   Signin: { screen: Signin },
+  Signin: { screen: Signin },
   NewGroup: { screen: NewGroup },
+  SearchGroup: { screen: SearchGroup },
+  Group: { screen: Group},
 }, {
   mode: 'modal',
 });
--- a/client/src/screens/group.screen.js
+++ b/client/src/screens/group.screen.js
@@ -1,0 +1,167 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import {
+  FlatList,
+  ActivityIndicator,
+  Button,
+  Text,
+  TouchableHighlight,
+  View,
+} from 'react-native';
+import { compose } from 'react-apollo';
+import moment from 'moment';
+import Icon from 'react-native-vector-icons/FontAwesome';
+import { connect } from 'react-redux';
+
+import { extendAppStyleSheet } from './style-sheet';
+
+const styles = extendAppStyleSheet({
+  groupContainer: {
+    flex: 1,
+    flexDirection: 'row',
+    alignItems: 'center',
+    backgroundColor: 'white',
+    borderBottomColor: '#eee',
+    borderBottomWidth: 1,
+    paddingHorizontal: 12,
+    paddingVertical: 8,
+  },
+  groupName: {
+    fontWeight: 'bold',
+    flex: 0.7,
+  },
+  groupTextContainer: {
+    flex: 1,
+    flexDirection: 'column',
+    paddingLeft: 6,
+  },
+  groupText: {
+    color: '#8c8c8c',
+  },
+  groupImage: {
+    width: 54,
+    height: 54,
+    borderRadius: 27,
+  },
+  groupTitleContainer: {
+    flexDirection: 'row',
+  },
+  groupLastUpdated: {
+    flex: 0.3,
+    color: '#8c8c8c',
+    fontSize: 11,
+    textAlign: 'right',
+  },
+  groupUsername: {
+    paddingVertical: 4,
+  },
+  header: {
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+  },
+});
+
+// format createdAt with moment
+const formatCreatedAt = createdAt => moment(createdAt).calendar(null, {
+  sameDay: '[Today]',
+  nextDay: '[Tomorrow]',
+  nextWeek: 'dddd',
+  lastDay: '[Yesterday]',
+  lastWeek: 'dddd',
+  sameElse: 'DD/MM/YYYY',
+});
+
+const Header = ({ onPress, onPressFind }) => (
+  <View style={styles.header}>
+  </View>
+);
+Header.propTypes = {
+};
+
+class Member extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { id, username } = this.props.member;
+    return (
+      <TouchableHighlight
+        key={id}
+      >
+        <View style={styles.groupContainer}>
+          <Icon name="user" size={24} color={'blue'} />
+          <View style={styles.groupTextContainer}>
+            <View style={styles.groupTitleContainer}>
+              <Text style={styles.groupName} numberOfLines={1}>{username}</Text>
+              <Text style={styles.groupLastUpdated}>{id}</Text>
+            </View>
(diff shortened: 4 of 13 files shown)


OSPFNeighbour:
Fixes #38

OSPFNeighbour:
work around for #22

client/src/graphql/all-groups.query.js
```diff
@@ -0,0 +1,29 @@
+import gql from 'graphql-tag';
+
+// get the user and all user's groups
+export default gql`
+query{
```
sdunster: indentation

client/src/graphql/all-groups.query.js
```diff
@@ -0,0 +1,29 @@
...
+user
+{
+  id
+  organisation
```
sdunster: please follow conventions:
```
  organisation {
    ...
```
not
```
  organisation
  {
    ...
```

client/src/graphql/join-group.mutation.js
```diff
@@ -0,0 +1,9 @@
+import gql from 'graphql-tag';
+
+export default gql`
+ mutation addUserToGroup($groupUpdate: AddUserToGroupInput!) {
```
sdunster: 2 spaces for indents

client/src/navigation.js
```diff
@@ -41,7 +44,10 @@ const MainScreenNavigator = TabNavigator({
 const AppNavigator = StackNavigator({
   Main: { screen: MainScreenNavigator },
   Signin: { screen: Signin },
+  Signin: { screen: Signin },
   NewGroup: { screen: NewGroup },
+  SearchGroup: { screen: SearchGroup},
```
sdunster: space before `}`

client/src/screens/group.screen.js
```diff
@@ -0,0 +1,210 @@
...
+};
+
+class Member extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { id, username } = this.props.member;
+    console.log(this.props)
```
sdunster: plz no console.log unless really worthwhile

client/src/screens/group.screen.js
```diff
@@ -0,0 +1,210 @@
...
+
+Member.propTypes = {
+  member: PropTypes.shape({
+    id: PropTypes.number.isRequired,
+    username: PropTypes.string.isRequired,
+  }),
+};
+
+class Group extends Component {
+  static navigationOptions = {
```
sdunster: uhh is this even used?

client/src/screens/group.screen.js
```diff
@@ -0,0 +1,210 @@
...
+    // render group stuff
+    return (
+      <View style={styles.container}>
+      <Text style={styles.warning}>{this.props.navigation.state.params.title}</Text>
+      <Text style={styles.warning}>{this.props.navigation.state.params.tags}</Text>
+
+      <FlatList
+        //TODO: Make a query that pulls just the one group back, not all of them
+        //find the array object with the same name as the one passed
+        data={user.organisation.groups.find((element) => {if(element.name == this.props.navigation.state.params.title){return element}}).users}
```
sdunster: this needs to go in a function.

client/src/screens/group.screen.js
```diff
@@ -0,0 +1,210 @@
...
+        keyExtractor={this.keyExtractor}
+        renderItem={this.renderItem}
+        ListHeaderComponent={() => <Header onPress={this.goToNewGroup} onPressFind={this.goToSearchGroup} />}
```
sdunster: no capital letter for parameters: `listHeaderComponent` not `ListHeaderComponent`

client/src/screens/groups.screen.js
```diff
@@ -70,24 +72,27 @@ const formatCreatedAt = createdAt => moment(createdAt).calendar(null, {
...
 class Group extends Component {
   constructor(props) {
     super(props);
-
     this.goToGroup = this.props.goToGroup.bind(this, this.props.group);
   }
 
   render() {
     const { id, name } = this.props.group;
+    const tags = (this.props.group.tags.map((elem) => {return '#'+elem.name})).join(',')
```
sdunster: function

client/src/screens/groups.screen.js
```diff
@@ -145,20 +155,27 @@ class Groups extends Component {
 
   goToGroup(group) {
     const { navigate } = this.props.navigation;
-    navigate('Group', { groupId: group.id, title: group.name });
+    const tags = (group.tags.map((elem) => {return '#'+elem.name})).join(',')
+
+    navigate('Group', { groupId: group.id, title: group.name, tags: tags });
```
sdunster: I don't think you want to put tags in the navigation params - the groupId and title should be enough. You can get the rest from within the components on that screen.

client/src/screens/groups.screen.js
```diff
@@ -170,8 +187,8 @@ class Groups extends Component {
     if (user && !user.groups.length) {
       return (
         <View style={styles.container}>
-          <Header onPress={this.goToNewGroup} />
-          <Text style={styles.warning}>{'You do not have any groups.'}</Text>
+          <Header onPress={this.goToNewGroup} onPressFind={this.goToSearchGroup}/>
+          <Text onPress={this.onRefresh} style={styles.warning}>{'You are not a member of any groups. click to reload'}</Text>
```
sdunster: we can do better than this

client/src/screens/groups.screen.js
```diff
@@ -183,7 +200,7 @@ class Groups extends Component {
           data={user.groups}
           keyExtractor={this.keyExtractor}
           renderItem={this.renderItem}
-          ListHeaderComponent={() => <Header onPress={this.goToNewGroup} />}
+          ListHeaderComponent={() => <Header onPress={this.goToNewGroup} onPressFind={this.goToSearchGroup} />}
```
sdunster: lowerCase plz

client/src/screens/search-groups.screen.js
```diff
@@ -0,0 +1,265 @@
...
+
+  }
+
+  render() {
+    const { id, name } = this.props.group;
+    console.log(this.myGroups)
+    const tags = (this.props.group.tags.map((elem) => {return '#'+elem.name})).join(',')
+    //color the already subscribed groups green
+    let iHaveDisAlready = false
+    this.myGroups.some(function(a) {
```
sdunster: couldn't you do something like this? you might need to use `_.any`
```
const memberAlready = this.myGroups.any(g => a.id == id);
```

client/src/screens/search-groups.screen.js
```diff
@@ -0,0 +1,265 @@
...
+    this.onRefresh = this.onRefresh.bind(this);
+  }
+
+  onRefresh() {
+    this.props.refetch();
+  }
+
+  keyExtractor = item => item.id;
+
+  renderItem = ({ item }) => <Group group={item} myGroups={this.props.user.groups} joinGroup={this.joinGroup} joinGroupQuery={this.props.groupUpdate}/>;
```
sdunster: break onto 2 lines. Put parens around it if necessary.

client/src/screens/search-groups.screen.js
```diff
@@ -0,0 +1,265 @@
...
+
+  joinGroup(group) {
+    this.props.joinGroupQuery({groupId:group.id}).then((result) => {
+      alert('joined - '+group.name)
```
sdunster: does this even work? We shouldn't need to show confirmation to users but we do need to make sure we display an error if it fails. See login/signup for inspiration.

client/src/screens/search-groups.screen.js
```diff
@@ -0,0 +1,265 @@
...
+        <View style={[styles.loading, styles.container]}>
+          <ActivityIndicator />
+        </View>
+      );
+    }
+
+    if (user && !user.organisation.groups.length) {
+      return (
+        <View style={styles.container}>
+          <Text style={styles.warning}>{'no groups in this org'}</Text>
```
sdunster: c'mon you can do better: "There are no groups in this organisation"

server/src/logic.js
```diff
@@ -225,21 +225,27 @@ export const groupHandler = {
...
     const { name } = args.group;
     return getAuthenticatedUser(ctx).then(user =>
-      Creators.group({ name, user }),
+      Creators.group({ name, user, organisation: 1 }), //TODO: dont be static
     );
   },
   addUserToGroup(_, args, ctx) {
-    return getAuthenticatedUser(ctx).then(() =>
-      Group.findById(args.groupUpdate.groupId).then((group) => {
+    const { userId,groupId } = args.groupUpdate;
```
sdunster: all of this stuff needs to be in a seperate pull request and come with tests imo

server/src/schema.js
```diff
@@ -39,7 +38,7 @@ export const Schema = [`
 
   input AddUserToGroupInput {
     groupId: Int!
-    userId: Int!
+    userId: Int
```
sdunster: why would this become optional?

client/src/screens/group.screen.js
OSPFNeighbour: It needs to not exist at all

client/src/screens/group.screen.js
OSPFNeighbour: I just used what was in the tutorial? weird that it would be wrong and still work.

client/src/screens/groups.screen.js
OSPFNeighbour: Why refetch data we already know?

client/src/screens/groups.screen.js
OSPFNeighbour: Refreshed, found no easy answer, this works for POC

client/src/screens/search-groups.screen.js
OSPFNeighbour: yes it works, long term we want to redraw the whole page to reflect the change but I couldnt work that out at the time and moved on to other more important things.

server/src/schema.js
OSPFNeighbour: Because the client doesn't know its client ID, so we shouldn't make it required to make this call. the server can do the lookup.

client/src/screens/group.screen.js
OSPFNeighbour: yes. thats the standard StackNavigator nav options static.

client/src/screens/groups.screen.js
OSPFNeighbour: how come? its the only placed its used, if I needed it twice I would.

client/src/screens/groups.screen.js
OSPFNeighbour: as per other one

client/src/screens/search-groups.screen.js
OSPFNeighbour: Possibly, but is there any reason to given what ive written is correct?

(short of making it ES6 with a fat arrow)

OSPFNeighbour:
join group

OSPFNeighbour:
Hey Sam,

fixed the things we discussed and changed the page logic. group screen makes no data call, it is pushed all the data from the previous page.

client/src/screens/search-groups.screen.js
sdunster: Mine is about 5 lines shorter and less complicated.

server/src/schema.js
sdunster: good call - we should remove it entirely.

sdunster: we can make it optional later if we want to support adding users other than ourselves.

sdunster:
I confirmed this worked on iOS as expected. Note that while group creation inserts the appropriate group data into the local cache, joining an existing group currently does not.
screen shot 2017-11-22 at 23 56 01

PR #44, opened 2017-11-23
Ensure signin screen state is initialised

### LINKED ISSUE #41: Signing in with no password doesn't currently work
The intention was that if you didn't provide a password we would just let you become that user anyway for testing. This doesn't work because the password is a required field in graphql.
Interestingly, if you type a character in the password field and then delete it, it passes. So this just needs to be a client-side change that initialises the password state to '' rather than null.

sdunster:
* Set the state of `username`, `email` and `password` to `''`.
* Fixes #41

### DIFF
--- a/client/src/screens/signin.screen.js
+++ b/client/src/screens/signin.screen.js
@@ -81,6 +81,9 @@ class Signin extends Component {
     super(props);
     this.state = {
       view: 'login',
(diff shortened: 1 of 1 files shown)


sdunster:
Verified functionality on iOS simulator. This should be standard across Android as well.

PR #47, opened 2017-11-23
Linter configuration/tweaks

sdunster:
* Split configurations to be separate for server vs client
* Use react/JSX stuff only on client
* Try to turn down some of the noise about dumb stuff

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -13,7 +13,11 @@ module.exports = {
       "skipShapeProps": true
     }],
     "react/no-multi-comp": [0],
-    "no-bitwise": [0],
+    "function-paren-newline": ["error", "consistent"],
+    "object-curly-newline": ["error", {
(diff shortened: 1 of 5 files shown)


client/.eslintrc.js
```diff
@@ -13,7 +13,11 @@ module.exports = {
       "skipShapeProps": true
     }],
     "react/no-multi-comp": [0],
-    "no-bitwise": [0],
+    "function-paren-newline": ["error", "consistent"],
```
OSPFNeighbour: Dont think this is working. I still get hundreds of

Unexpected newline before ')'. (function-paren-newline)

OSPFNeighbour: wait, ignore me. it should be throwing as an error

sdunster:
#66 fixes all lint errors using this configuration.

PR #49, opened 2017-11-23
Better creator validation and fix mutation bugs

sdunster:
* Creators now validate that the entities that are passed into them have
  the `id` field which would otherwise result in a silent failure to actually
  associate things in the DB.
* Fix bugs:
  * When creating a group the creating user is not added to it.
  * When creating a schedule the group it is created for is not associated to it


sdunster:
This is actually untested currently but I'll test most of it while working on the groups stuff...

sdunster:
Confirmed this works with #37

Issue #50, opened 2017-11-23
Optimistically update the local cache when the user joins a group


sdunster:
Just like we do when we create a new group: https://github.com/NSWSESMembers/availability-poc/blob/master/client/src/screens/new-group.screen.js#L180

sdunster:
Closed via #61

PR #58, opened 2017-11-24
Optimistically update the local cache when the user joins a group

OSPFNeighbour:
update group search and join screens to use Graphql caching.

Im stuck on the following with an undefined variable.

```
const joinGroupQuery = graphql(JOIN_GROUP_QUERY, {
  props: ({ ownProps, mutate }) => ({
    groupUpdate: ({ groupId }) =>
      mutate({
        variables: { groupUpdate: { groupId } },
        update: (store, { data: { groupUpdate } }) => {
          const data = store.readQuery({
            query: ALL_GROUPS_QUERY,
            variables: { id: ownProps.auth.token },
          });
          data.user.groups.push(groupUpdate);
          store.writeQuery({
            query: ALL_GROUPS_QUERY,
            variables: { id: ownProps.auth.token },
            data,
          });
        },
      }),
  }),
});
```

data.user.groups.push(groupUpdate); - groupUpdate is undefined.

I can see data is coming back as i have a then() on the call which contains the new group created buy for some reason I cant get it inside the update call.

### DIFF
--- a/client/src/graphql/join-group.mutation.js
+++ b/client/src/graphql/join-group.mutation.js
@@ -4,6 +4,7 @@ export default gql`
   mutation addUserToGroup($groupUpdate: AddUserToGroupInput!) {
     addUserToGroup(groupUpdate: $groupUpdate) {
       id
+      name
     }
   }
 `;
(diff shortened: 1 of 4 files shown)


sdunster:
Tim and I talked on Discord - we have a solution.

sdunster:
This is for #50

PR #59, opened 2017-11-24
Tdykes disable codepush in dev

OSPFNeighbour:
disabled codepush in dev mode via __DEV__ check and setting manual check vs auto check

### DIFF
--- a/client/index.android.js
+++ b/client/index.android.js
@@ -4,13 +4,13 @@ import {
 import codePush from 'react-native-code-push';
 import App from './src/app';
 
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
Tested the DEV side but not yet tested the production side.

sdunster:
Can you find out in which circumstances `__DEV__` is `true` or `false`? There's a mix of different parameters that might affect it.

OSPFNeighbour:
when ever the packer is running. can be overridden with ```--dev false```

PR #60, opened 2017-11-24
Fixes bad filter used to update graphQL cache on new group page

OSPFNeighbour:
was using a variable that didnt exist

### DIFF
--- a/client/src/screens/new-group.screen.js
+++ b/client/src/screens/new-group.screen.js
@@ -173,7 +173,7 @@ FinalizeGroup.propTypes = {
 };
 
 const createGroupMutation = graphql(CREATE_GROUP_MUTATION, {
-  props: ({ ownProps, mutate }) => ({
+  props: ({ mutate }) => ({
     createGroup: ({ name }) =>
       mutate({
         variables: { group: { name } },
@@ -181,16 +181,14 @@ const createGroupMutation = graphql(CREATE_GROUP_MUTATION, {
           // Read the data from our cache for this query.
           const data = store.readQuery({
             query: CURRENT_USER_QUERY,
-            variables: { id: ownProps.auth.id },
           });
 
-          // Add our message from the mutation to the end.
+          // Add our group from the mutation to the end.
           data.user.groups.push(createGroup);
 
           // Write our data back to the cache.
           store.writeQuery({
             query: CURRENT_USER_QUERY,
-            variables: { id: ownProps.auth.id },
             data,
           });
         },


client/src/screens/new-group.screen.js
```diff
@@ -181,16 +181,16 @@ const createGroupMutation = graphql(CREATE_GROUP_MUTATION, {
           // Read the data from our cache for this query.
           const data = store.readQuery({
             query: CURRENT_USER_QUERY,
-            variables: { id: ownProps.auth.id },
+            variables: { username: ownProps.auth.username },
```
sdunster: ooh actually we can remove this entirely - we don't pass any variables to `current-user.query`.

client/src/screens/search-groups.screen.js
```diff
@@ -68,21 +69,37 @@ class Group extends Component {
   constructor(props) {
     super(props);
     this.myGroups = this.props.myGroups;
-    this.joinGroupQuery = this.props.joinGroupQuery.bind(this);
-    this.joinGroup = this.props.joinGroup.bind(this, this.props.group);
+    this.joinGroup = this.joinGroup.bind(this);
+    this.groupUpdate = this.props.groupUpdate.bind(this)
+    this.state = {
+      memberAlready: _.some(this.myGroups, g => g.id === this.props.group.id),
```
sdunster: nice!

client/src/screens/search-groups.screen.js
```diff
@@ -68,21 +69,37 @@ class Group extends Component {
...
+    };
+
+  }
+
+  joinGroup() {
+    const { groupUpdate, group } = this.props;
+    groupUpdate({ groupId: group.id }).then((res) => {
+      this.setState({memberAlready:true });
```
sdunster: nice - but lets make sure this is actually needed - if we do it right we shouldn't need this because the apollo store update should cause a refresh (unchecked hypothesis)

client/src/screens/search-groups.screen.js
```diff
@@ -68,21 +69,37 @@ class Group extends Component {
...
+    }).catch((error) => {
+      Alert.alert(
+        'Error Joing Group',
```
sdunster: s/Joing/Joining

client/src/screens/search-groups.screen.js
sdunster: there is one downside to this approach, though... this is only checked at start time which means that if something updates the data and calls `render()` this state will not be updated. For that reason it probably should be moved back to `render()`

client/src/screens/search-groups.screen.js
```diff
@@ -212,29 +221,38 @@ AllGroups.propTypes = {
...
-        ),
       }),
     ),
   }),
 };
 
 const groupsQuery = graphql(ALL_GROUPS_QUERY, {
   skip: ownProps => !ownProps.auth || !ownProps.auth.token,
-  props: ({ data: { loading, networkStatus, refetch, user } }) => ({
+  props: ({ data: { loading, networkStatus, refetch, user }}) => ({
```
sdunster: the space at the end of an object literal is required per airbnb JS style

client/src/screens/search-groups.screen.js
OSPFNeighbour: It cases the group object to re-render but it wont have any new data because the AllGroup object isnt re-rendering, so it doenst know its joined a new group

client/src/screens/new-group.screen.js
OSPFNeighbour: Aah right. my bad.

OSPFNeighbour:
I am the GIT Master!

PR #61, opened 2017-11-24
Mutation query cache for Groups

OSPFNeighbour:
Added update function to groups mutation to enable predictive UI for joining a group and viewing joined groups.

group cache

### DIFF
--- a/client/src/screens/search-groups.screen.js
+++ b/client/src/screens/search-groups.screen.js
@@ -15,8 +15,7 @@ import _ from 'lodash';
 
 import { extendAppStyleSheet } from './style-sheet';
 import ALL_GROUPS_QUERY from '../graphql/all-groups.query';
-import JOIN_GROUP_QUERY from '../graphql/join-group.mutation';
-
+import JOIN_GROUP_MUTATION from '../graphql/join-group.mutation';
 
 const styles = extendAppStyleSheet({
   groupContainer: {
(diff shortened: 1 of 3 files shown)


OSPFNeighbour:
hub is showing a bunch of rebases as new commits. there not, click one and you will see it was a rebase!

OSPFNeighbour:
Fixes #50

OSPFNeighbour:
Answered why the other query is also getting updated;

This option allows you to update your store based on your mutation’s result. By default Apollo Client will update all of the overlapping nodes in your store. Anything that shares the same id as returned by the dataIdFromObject you defined will be updated with the new fields from your mutation results. However, sometimes this alone is not sufficient. Sometimes you may want to update your cache in a way that is dependent on the data currently in your cache. For these updates you may use an options.update function.

OSPFNeighbour:
@sdunster have a look now please. I worked out how to do this without the hacky redraw.

Flatlist was the thing killing it, its PURE so wont update in the normal fashion like other elements do

client/src/screens/search-groups.screen.js
```diff
@@ -68,21 +67,36 @@ class Group extends Component {
...
 
+
   render() {
     const { id, name } = this.props.group;
     const tags = this.props.group.tags.map(elem => `#${elem.name}`).join(',');
-
-    // color the already subscribed groups green
-    const memberAlready = _.some(this.myGroups, g => g.id === id);
-
+    this.state = {
```
sdunster: hmm I don't think you should store this in state.. why didn't what we had previous work? (using a local)
If you're going to go and calculate the state every time you render and the only place you use it is the render function it seems silly to store it in state.

sdunster:
No it looks like what you're doing with `FlatList` is right as far as tricking it to reload by passing `extraData` goes. State thing should probably still be addressed though.

PR #62, opened 2017-11-24
fix object type in prototype on group page to be object not string

OSPFNeighbour:
Dunno how we didnt notice that.


client/src/screens/group.screen.js
```diff
@@ -129,7 +129,7 @@ Group.propTypes = {
   navigation: PropTypes.shape({
     state: PropTypes.shape({
       params: PropTypes.shape({
-        group: PropTypes.string.isRequired,
+        group: PropTypes.object,
```
sdunster: use shape

OSPFNeighbour:
Fixed as requested

PR #67, opened 2017-11-25
Travis now does linting for both client and server

sdunster:
* Implement travis matrix testing so that different components are tested
  in different jobs.
* Create new shell script to handle install, build and testing.
* When testing server we also now handle running the server in the background.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,8 @@
 language: node_js
 node_js: 6
-cache:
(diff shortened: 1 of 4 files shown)


sdunster:
I pushed a broken test to confirm that the existing tests still work.

sdunster:
This won't pass until #47 and #66 land.

PR #68, opened 2017-11-27
Default to using local GraphQL server if __DEV__

### LINKED ISSUE #63: Automatically switch to using local GraphQL server when `__DEV__`

sdunster:
* When `__DEV__` is true (development builds of the app/react native bundler is running) we default to using `http://localhost:8080/graphql`.
* This might not work for everyone but that is OK - they can change it.
* If someone accidentally commits their local IP or whatever it won't affect production builds.
* Closes #63

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -6,6 +6,9 @@ module.exports = {
     "jsx-a11y",
     "import"
   ],
+  "globals": {
+    "__DEV__": false, // enforce immutability
+  },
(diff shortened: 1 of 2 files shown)


sdunster:
Confirmed my iOS simulator app was hitting my local GraphQL server.

OSPFNeighbour:
FYI this doesnt work on Android emulators, Nor will it work on physical android debugs.

PR #71, opened 2017-11-27
Client package upgrades/modernisation

sdunster:
* Held back the following:
  * apollo-client - still beta
  * react-apollo - depends on new version of apollo-client
  * react - react-native wants 16.0.0 specifically
* Adopt prop-types because React v16 will require it
* Adopt new redux-persist API
* Update project files to match react-native changes.
  * Did not adopt unified index.js - we'll do this later so as not to break code push.
* More error checking in local state reducers.
* Remove `redux-devtools-extension` since it is designed for browser use.
  * We could add `remote-redux-devtools` and `remotedev-rn-debugger` later.


OSPFNeighbour:
LGTM

PR #72, opened 2017-11-27
logic for returning user schedules, further test data, testes to test…

OSPFNeighbour:
Added resolver for user -> schedules.

Added some more test data to help test the scope of the return.

Added jest tests to confirm schedules are returned.

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -6,6 +6,9 @@ module.exports = {
     "jsx-a11y",
     "import"
   ],
+  "globals": {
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
@sdunster can you check my sequalise query, Its relying on some internal magic to only return groups where there is a user relationship. If i try force that with another include it gets cranky.

Confirmed this by making more schedules in groups the test users isnt in, and they are not returned.

sdunster:
You sure all the code is here? All I see is that you're importing sequelize for no apparent reason.

PR #75, opened 2017-11-27
Created logic for returning user schedules.

OSPFNeighbour:
Created logic to return users schedules.

Created more test data to support this.

Created jest tests to test responses.

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -1,8 +1,9 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
 
+
 import JWT_SECRET from './config';
-import { Group, User, Device, Organisation } from './models';
+import { Group, User, Device, Organisation, Schedule } from './models';
 import Creators from './creators';
 
 // reusable function to check for a user with context
@@ -60,8 +61,11 @@ export const userHandler = {
     return user.getEvents();
   },
   schedules() {
-    // TODO: implement this
-    return [];
+    return Schedule.findAll({
+      include: [
+        { model: Group, required: true },
+      ],
+    });
   },
   organisation(user) {
     return user.getOrganisation();
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -60,11 +60,10 @@ export const userHandler = {
   events(user) {
     return user.getEvents();
   },
-  schedules() {
+  schedules(user) {
+    const groups = user.getGroups();
     return Schedule.findAll({
-      include: [
-        { model: Group, required: true },
-      ],
+      where: { groupId: { $in: groups.map(g => g.id) } },
     });
   },
   organisation(user) {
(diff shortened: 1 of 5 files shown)


server/src/logic.js
```diff
@@ -60,8 +61,11 @@ export const userHandler = {
     return user.getEvents();
   },
   schedules() {
-    // TODO: implement this
-    return [];
+    return Schedule.findAll({
+      include: [
+        { model: Group, required: true },
```
sdunster: this doesn't actually filter by user though. So every time we resolve this for any user we will return all schedules from all groups (it's not even limited by organisation!)
If you're just looking to stuff data in the API this is fine but lets make a note of it with a comment.

server/src/test-data.js
```diff
@@ -11,7 +11,15 @@ export const loadTestData = () =>
         password: '123',
         email: 'test@test.com',
         organisation,
-      }),
+      }).then(user =>
+        Promise.all([
```
sdunster: this `.all()` is redundant but you probably know that

server/tests/current-user.test.js
```diff
@@ -98,20 +98,25 @@ describe('GraphQL query - Current user', () => {
           id
           schedules {
             id
-            timeSegments {
```
sdunster: sure you want to get rid of this?

server/src/test-data.js
```diff
@@ -56,11 +64,6 @@ export const loadTestData = () =>
             user,
             organisation,
           }),
-          Creators.group({
```
sdunster: intentional remove?

server/src/test-data.js
OSPFNeighbour: Later down the road i will need further things in here, so this got it inline with the other test users

server/src/logic.js
OSPFNeighbour: You sure? in my testing it seems to only return things that user had a relation to. I created another set of groups that this user was not in and they are not returned

server/tests/current-user.test.js
OSPFNeighbour: yes because we cant resolve that yet, so its basically testing for something that wont work

server/src/logic.js
```diff
@@ -59,9 +60,11 @@ export const userHandler = {
   events(user) {
     return user.getEvents();
   },
-  schedules() {
-    // TODO: implement this
-    return [];
+  schedules(user) {
+    const groups = user.getGroups();
+    return Schedule.findAll({
+      where: { groupId: { $in: groups.map(g => g.id) } },
```
sdunster: perfect!

OSPFNeighbour:
Or we event want to be able to go user -> schedules or do we want to force user -> groups -> schedules? if we dont want this short cut then this PR can be killed.

PR #76, opened 2017-11-27
Basic UI for viewing events

sdunster:
* Uses client-mocked data for now but should be pretty easy to swap in with stuff from server.
* This view is designed to be view-only for users who want to see what is happening with an event but aren't necessarily responding to or otherwise interacting with it. I suspect we will build a seperate responder interface that allows updates of your responding status.

### DIFF
--- a/client/src/screens/event-detail.screen.js
+++ b/client/src/screens/event-detail.screen.js
@@ -1,0 +1,254 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import {
+  FlatList,
+  ActivityIndicator,
+  Text,
+  View,
+} from 'react-native';
+import { graphql, compose } from 'react-apollo';
+import Icon from 'react-native-vector-icons/FontAwesome';
+import { connect } from 'react-redux';
+import _ from 'lodash';
+
+import { extendAppStyleSheet } from './style-sheet';
+import CURRENT_USER_QUERY from '../graphql/current-user.query';
+
+const styles = extendAppStyleSheet({
+  respondContainer: {
+    flex: 1,
+    flexDirection: 'row',
+    alignItems: 'center',
+    backgroundColor: 'white',
+    borderBottomColor: '#eee',
+    borderBottomWidth: 1,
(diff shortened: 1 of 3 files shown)


sdunster:
This PR is based off of #71 and #74 so I suggest you only review the latest commit: 9d089ba

Tested only on iOS:
screen shot 2017-11-26 at 22 24 09
screen shot 2017-11-26 at 22 24 11

Work still to be done:
* Replace MAP placeholder with actual map. Possibly this: https://github.com/airbnb/react-native-maps
* Hook it up to the actual API rather than use mocks. We might need to re-jigger the fields a bit since the API is pretty bare atm.
* Decide what metadata we actually want to show.
* Set up subscriptions so this updates in real time.

sdunster:
#71 and #74 landed - rebased onto master.

client/src/screens/event-detail.screen.js
```diff
@@ -0,0 +1,254 @@
...
+    <View style={styles.respondContainer}>
+      <Icon name="user" size={24} color={color} />
+      <View style={styles.respondTextContainer}>
+        <Text style={styles.respondName} numberOfLines={1}>{user.name}</Text>
+        <Text style={styles.respondStatus} numberOfLines={1}>{status} - {detail}</Text>
+      </View>
+    </View>
+  );
+};
+EventResponse.propTypes = {
```
OSPFNeighbour: probably all isRequired

client/src/screens/event-detail.screen.js
```diff
@@ -0,0 +1,254 @@
...
+          renderItem={this.renderItem}
+          onRefresh={this.onRefresh}
+          extraData={event} // redraw if this changes
+          refreshing={networkStatus === 4}
+        />
+      </View>
+    );
+  }
+}
+EventDetail.propTypes = {
```
OSPFNeighbour: again I would suggest some of these are ```isRequired``` to catch malformed data

OSPFNeighbour:
Seems about right.

PR #77, opened 2017-11-27
Addition of Dumb Components

sjmcculloch:
-- Add new "Components" folder with an initial List Component
-- Moved datasources out of home.screen.js to new /fixtures folder

The Components folder is to hold dumb components. These hold no state, receive data through props and handle events through callbacks to their containers (screens). It also helps to maintain styles within the components rather than a more traditional large stylesheet across the app. The goal is to have a component library like the shown here:

https://airbnb.design/wp-content/uploads/2016/04/Organized-airbnb-dls-components-sample.png

For a full explanation of the smart vs dumb components, see here:

https://jaketrent.com/post/smart-dumb-components-react/

I have some more fleshed out examples here:

https://github.com/sjmcculloch/CurrencyConverter/tree/master/app/components


client/src/fixtures/availability.js
```diff
@@ -0,0 +1,16 @@
+const availability = [
+    {
```
sdunster: double indent

client/src/fixtures/events.js
```diff
@@ -0,0 +1,18 @@
+const events = [
+    {
```
sdunster: double indent

sdunster:
Lint errors are here if you're too lazy to run `yarn lint`: https://travis-ci.org/NSWSESMembers/availability-poc/jobs/308161911

sjmcculloch:
Ah - I'll fix those - I normally just use eslint inside vscode and do it manually.

sjmcculloch:
This should be good to go now @sdunster - let me know if I missed anything.

PR #78, opened 2017-11-28
Updated resolvers & logic to support events -> responses.

OSPFNeighbour:
Created logic paths to support returning events and responses under user.

Creators to create event response
logic to handle elements returned for event and responses
model to have event response model, and relationships
schema to handle data return
test-data to populate an event response for test user
just test to test the above

Question:

Do we want user -> events -> responses to show all responses or just current users?

I would suggest it only show current users answers, as it lives under user? But I can see the argument to show all users responses for cache reasons. Possibly put under user -> org -> events everyones responses? This question needs to be answered for schedules as well as it has the same sorta scope. I have coded this to show all users responses. Code assumes user is in the group that the event is attached to, we need to do some re-jig to change that logic, possibly expose this under user -> org -> events if we want non member users to see it.

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -1,7 +1,8 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
 import JWT_SECRET from './config';
-import { Group,
+import {
+  Group,
   User,
   Device,
   Organisation,
@@ -220,8 +221,8 @@ export const scheduleHandler = {
 };
 
 export const eventHandler = {
-  group(eventt) {
-    return eventt.getGroup();
+  group(event) {
+    return event.getGroup();
   },
   responses(event) {
     return EventResponse.findAll({
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -1,9 +1,15 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
-
-
 import JWT_SECRET from './config';
-import { Group, User, Device, Organisation, Schedule, TimeSegment } from './models';
+import { Group,
+  User,
+  Device,
+  Organisation,
+  Schedule,
+  TimeSegment,
+  Event,
+  EventResponse,
+} from './models';
 import Creators from './creators';
 
 // reusable function to check for a user with context
@@ -58,7 +64,10 @@ export const userHandler = {
     return user.getGroups();
   },
   events(user) {
-    return user.getEvents();
+    return user.getGroups()
+      .then(groups => Event.findAll({
+        where: { groupId: groups.map(g => g.id) },
+      }));
   },
   schedules(user) {
     return user.getGroups()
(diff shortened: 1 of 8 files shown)


sdunster:
I pretty firmly believe the API should return all responses, especially because the resolver should be generic and not do something specific only when it detects this path (user -> events -> responses). But we could add a parameter that filters it so that it only returns the current user's responses or even filter by user ID or something.

server/src/logic.js
```diff
@@ -1,9 +1,15 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
-
-
 import JWT_SECRET from './config';
-import { Group, User, Device, Organisation, Schedule, TimeSegment } from './models';
+import { Group,
```
sdunster: if you're going to layout like this put the first one on a new line

server/src/logic.js
```diff
@@ -210,6 +219,23 @@ export const scheduleHandler = {
   },
 };
 
+export const eventHandler = {
+  group(eventt) {
```
sdunster: typo?

server/src/logic.js
```diff
@@ -210,6 +219,23 @@ export const scheduleHandler = {
...
+    return eventt.getGroup();
+  },
+  responses(event) {
+    return EventResponse.findAll({
+      where: { EventId: event.id },
```
sdunster: i would normally expect `EventId` to be `eventId`

server/src/schema.js
```diff
@@ -101,6 +101,7 @@ export const Schema = [`
   }
 
   type Event {
+    group: Group!
```
sdunster: put this after `user` maybe? It looks out of place as the first thing - especially because it's before `id`.

server/src/test-data.js
```diff
@@ -105,7 +105,16 @@ export const loadTestData = () =>
               name: 'VR Wollongong Escarpment',
               details: 'VR operators required to assist with rescue',
               group,
-            }),
+            }).then(event =>
+              Creators.eventResponse({
+                status: 'On Route',
```
sdunster: didn't you make this field an int?

server/src/test-data.js
```diff
@@ -105,7 +105,16 @@ export const loadTestData = () =>
...
+                detail: 'Attending',
+                destination: 'Scene',
+                eta: 600,
```
sdunster: imo you should always use a timestamp for this. 600 seconds since what? record creation?

server/src/logic.js
OSPFNeighbour: wth, why did that even work.

server/src/test-data.js
OSPFNeighbour: incorrectly yes, I have changed the model to be a string (for now anyway, maybe we do want ints?)

sdunster: yeah eventually but str is fine for now - easy to change later

PR #85, opened 2017-11-28
Update Schedules Logic to support TimeSegments

OSPFNeighbour:
Schedules resolves can now return TimeSegements.

Added test data for this.
Added tests to test the test data.

### DIFF
--- a/server/src/test-data.js
+++ b/server/src/test-data.js
@@ -85,7 +85,8 @@ export const loadTestData = () =>
                 endTime: 1511854428,
                 schedule,
                 user,
-              })),
+              }),
+            ),
             Creators.schedule({
               group,
               name: 'Wollongong VR',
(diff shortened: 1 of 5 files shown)


server/src/test-data.js
```diff
@@ -78,7 +78,14 @@ export const loadTestData = () =>
...
               details: 'Weekly storm availability',
-            }),
+            }).then(schedule =>
+              Creators.timeSegment({
+                status: 'Available',
+                startTime: 1511853428,
+                endTime: 1511854428,
+                schedule,
+                user,
+              })),
```
sdunster: really unimportant nit (mostly feedback for next time rather than you should bother changing this):
put the last paren on this line on the next line so it is closed at the same indentation level as the one it is opened (L81). To pass lint you'll also need to an extra comma. The only reason I point this out is that my magical indentation/block parsing spidey-sense trips because I see an outdent with no closing paren.
L123-L124 are an example of how this should look.

PR #88, opened 2017-11-29
Tdykes.query.for.group.by.id

OSPFNeighbour:
Add schema to allow group query with id filter.
Add logic to handle passed Id's
Add jist tests to test both org and user groups.

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -59,7 +59,10 @@ export const userHandler = {
   query(_, __, ctx) {
     return getAuthenticatedUser(ctx);
   },
-  groups(user) {
+  groups(user, args) {
+    if (args.id) {
+      return user.getGroups({ where: { Id: args.id } });
+    }
     return user.getGroups();
   },
   events(user) {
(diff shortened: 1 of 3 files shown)


server/src/logic.js
```diff
@@ -59,7 +59,10 @@ export const userHandler = {
   query(_, __, ctx) {
     return getAuthenticatedUser(ctx);
   },
-  groups(user) {
+  groups(user, args) {
+    if (args.id) {
```
sdunster: Let it be known that I don't think this is the right place for this - it belongs as a root-level query but we can remove it later once we have an alternative

server/src/logic.js
```diff
@@ -59,7 +59,10 @@ export const userHandler = {
...
+      return user.getGroups({ where: { Id: args.id } });
```
sdunster: `Id` should be `id` (lower case)

server/src/logic.js
```diff
@@ -241,8 +244,11 @@ export const organisationHandler = {
     const { name } = args.organisation;
     return Creators.organisation({ name });
   },
-  groups(org) {
+  groups(org, args) {
     // TODO: think about who we show the complete organisation group list to
+    if (args.id) {
```
sdunster: as above

PR #89, opened 2017-11-30
Add react-native-maps component

sdunster:
* `react-native link react-native-maps`
* Drop it into event view to prove it works

Confirmed it works on iOS:
screen shot 2017-11-29 at 21 11 30

Need verification on Android. Might need Google maps API keys.


OSPFNeighbour:
Android:

24271554_10156187360693836_572571185_o

PR #90, opened 2017-12-03
Rebuild app, rename Callout

sdunster:
* Deleted ios and android directories
* Re-built from scratch with `react-native eject`
* Merged `index.ios.js` and `index.android.js` into `index.js` as per latest React Native recommendations.
* Re-linked plugins and configured:
  * `react-native-maps`
  * `react-native-vector-icons`
  * Code Push
  * Geolocation
* Added some really shitty placeholder icons
* App now named 'Callout' for now


sdunster:
Built and deployed both Android and iOS apps and both appear to work.
Tested:
* geolocation
* maps
* sign in

Push notifications might need additional configuration but we haven't used that yet anyway so I don't see this as a regression.

client/app.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "availpoc",
-  "displayName": "Avail",
+  "name": "Callout",
```
sdunster: this was always supposed to be `ProperCase` according to the docs

client/android/app/src/main/java/com/callout/MainApplication.java
```diff
@@ -33,10 +33,15 @@ public boolean getUseDeveloperSupport() {
       return Arrays.<ReactPackage>asList(
           new MainReactPackage(),
             new MapsPackage(),
-            new CodePush("ihP3wosJcBg8HofvREZAnBuaTzllae9e3d88-2347-4e09-a02a-76e4d3554374", getApplicationContext(), BuildConfig.DEBUG),
+            new CodePush(getResources().getString(R.string.reactNativeCodePush_androidDeploymentKey), getApplicationContext(), BuildConfig.DEBUG),
```
sdunster: it moved this to strings.xml which is nice

client/ios/Callout/Info.plist
```diff
@@ -37,8 +37,14 @@
 			</dict>
 		</dict>
 	</dict>
+	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
```
sdunster: it wasn't clear which ones of these were required so I added all of them. We're supporting background location updates in the shell, now.

client/ios/Callout/Info.plist
```diff
@@ -53,12 +59,15 @@
 		<string>SimpleLineIcons.ttf</string>
 		<string>Zocial.ttf</string>
 	</array>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>location</string>
```
sdunster: as per previous comment

client/ios/availpoc/AppDelegate.m
```diff
@@ -1,72 +0,0 @@
...
-
-// Required to register for notifications
-- (void)application:(UIApplication *)application didRegisterUserNotificationSettings:(UIUserNotificationSettings *)notificationSettings
-{
-  [RCTPushNotificationManager didRegisterUserNotificationSettings:notificationSettings];
-}
-// Required for the register event.
-- (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
-{
-  [RCTPushNotificationManager didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
```
sdunster: yeah this is missing from the new one so no push notifications but we can add this back in easily.

client/ios/availpoc/availpoc.entitlements
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-<plist version="1.0">
-<dict>
-	<key>aps-environment</key>
```
sdunster: we will need to make sure this is re-added when we set up push as well

sdunster:
On iOS we request background location updates. The intention is to be able to continue sharing your location on your way to HQ even if you're in another app. We will probably need to make a NUX to explain this to users and also put some thought into making sure we don't share their location for too long because it would drain the user's battery.

screen shot 2017-12-02 at 11 30 04

PR #96, opened 2017-12-03
Mutation to remove a user from passed group id

OSPFNeighbour:
Adds mutation to remove user from group by passed group id.

refactor group tests into seperate group test file.

Adds tests to remove and join a group.

### DIFF
--- a/server/tests/groups.test.js
+++ b/server/tests/groups.test.js
@@ -1,0 +1,64 @@
+import { itReturnsSuccess, run } from './common';
+
+describe('Manipulate user group membership - leave', () => {
+  const response = run(
+    {
+      query: `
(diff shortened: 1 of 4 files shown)


server/tests/groups.test.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+  );
+
+  itReturnsSuccess(response);
+  it('should return true', () => response.then((res) => {
+    expect(res.data.removeUserFromGroup).toEqual(true);
+  }));
+});
+
+describe('Manipulate user group membership - join', () => {
```
sdunster: if these tests are failing sporadically you probably just need to put them into a single describe() so that they run together. Make sure you leave group after you've joined it, obviously.

PR #97, opened 2017-12-03
Client - group details page rewrite

OSPFNeighbour:
Group details are pulled when the page is visited.

Page linking is fixed so everything points here.

Join/Leave button is smart enough to join or leave (pending #96)

### DIFF
--- a/client/src/screens/group.screen.js
+++ b/client/src/screens/group.screen.js
@@ -3,13 +3,22 @@ import React, { Component } from 'react';
 import {
   FlatList,
   Text,
+  ActivityIndicator,
   TouchableHighlight,
   View,
+  Button,
+  Alert,
 } from 'react-native';
+import { graphql, compose } from 'react-apollo';
 import Icon from 'react-native-vector-icons/FontAwesome';
 import { connect } from 'react-redux';
+import _ from 'lodash';
 
 import { extendAppStyleSheet } from './style-sheet';
+import SEARCH_GROUP_QUERY from '../graphql/search-group.query';
+import JOIN_GROUP_MUTATION from '../graphql/join-group.mutation';
+import LEAVE_GROUP_MUTATION from '../graphql/leave-group.mutation';
+import CURRENT_USER_QUERY from '../graphql/current-user.query';
 
 const styles = extendAppStyleSheet({
   groupContainer: {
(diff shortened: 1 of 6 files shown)


OSPFNeighbour:
So heres a question. how do we want to handle logic that removes data.

In the above i just return a bool for the deletion, and from that I edit the apollo cache. but because i dont return an object ID I cant update the cache across all queries, only the ones I open.

We could get the remove from group mutation to return the group object from which the user was removed, that would allow the cache to be updated easier.

sdunster:
Screenshots?

Yeah you answered your own question - the mutation should return everything you need to be able to update the cache.

OSPFNeighbour:
Thinking about it, I think a bool returned is enough. Doesn't make cache cleanup any easier if we return more.

OSPFNeighbour:
group screen

client/src/screens/group.screen.js
```diff
@@ -94,41 +178,131 @@ class Group extends Component {
...
+          error.message,
+          [
+            { text: 'OK', onPress: () => {} },
+          ],
+        );
+      });
+    } else {
+      leaveGroup({ groupId: group.id }).then(() => {
+        // pull group details down again to populate the group details
+        refetch();
```
sdunster: the fact you have to do this probably indicates something isn't set up quite right but it's probably not worth the time investment to investigate right now.

client/src/screens/group.screen.js
```diff
@@ -94,41 +178,131 @@ class Group extends Component {
...
+  onRefresh() {
+    this.props.refetch();
+  }
 
-  renderItem = ({ item }) => <Member member={item} />
+  joinOrLeaveGroup(group, memberAlready) {
+    const { joinGroup, leaveGroup, refetch } = this.props;
+    if (!memberAlready) {
+      joinGroup({ groupId: group.id }).then(() => {
+      // dont need to do anything, but we might in the future
```
sdunster: I understand why you put this comment here but I think it's probably best to leave it out - we can add the `.then()` in the future if we need it.

client/src/screens/search-groups.screen.js
```diff
@@ -94,7 +81,7 @@ class Group extends Component {
       memberAlready: _.some(this.props.myGroups, g => g.id === this.props.group.id),
     };
     return (
-      <TouchableHighlight key={id} onPress={this.joinGroup}>
+      <TouchableHighlight key={id} onPress={() => this.goToGroup(id)}>
```
sdunster: don't pass the `id` to the function here. Use what you had previously (get the group directly from `this.props.group.id`).

client/src/screens/search-groups.screen.js
```diff
@@ -67,23 +65,12 @@ class Group extends Component {
   constructor(props) {
     super(props);
     this.myGroups = this.props.myGroups;
-    this.joinGroup = this.joinGroup.bind(this);
-    this.groupUpdate = this.props.groupUpdate.bind(this);
+    this.navigation = this.props.navigation;
```
sdunster: bind goToGroup:
```
this.goToGroup = this.goToGroup.bind(this);
```

client/src/screens/search-groups.screen.js
sdunster: no anon func:
```
<TouchableHighlight key={id} onPress={this.goToGroup}>
```

server/src/logic.js
```diff
@@ -291,6 +291,20 @@ export const groupHandler = {
       }),
     );
   },
+  removeUserFromGroup(_, args, ctx) {
+    const { groupId } = args.groupUpdate;
+    return getAuthenticatedUser(ctx).then(user =>
+      Group.findById(groupId).then((group) => {
```
sdunster: we'll obviously need to do some permissions checks here but dw about it right now

server/src/logic.js
```diff
@@ -291,6 +291,20 @@ export const groupHandler = {
...
+        if (!group) {
+          return Promise.reject(Error('Invalid group!'));
+        }
+        return group.removeUser(user).then((rows) => {
```
sdunster: any time you return true/false inside an if statement like this you can always shorten :)
```
return group.removeUser(user).then(rows => !!rows);
```

client/src/screens/group.screen.js
OSPFNeighbour: I can either return the whole group when you leave and pass that back to the page (which would need a new query in the remove function because it returns a int of the number of rows removed, not an object), or trigger the page to refresh which will fetch the group again and update the cache by ID's for all linked items. Both ways need a query, this way is just easier on the cache

server/src/logic.js
OSPFNeighbour: Please try yank out e5f092d :-)

client/src/screens/search-groups.screen.js
OSPFNeighbour: how come? goToGroup is a function not a prop

OSPFNeighbour: dw, answered my own question

client/src/screens/search-groups.screen.js
OSPFNeighbour: done

OSPFNeighbour: done

client/src/screens/group.screen.js
```diff
@@ -70,22 +109,67 @@ const Member = (props) => {
...
+  );
+};
+ScheduleDisplay.propTypes = {
+  schedule: PropTypes.shape({
+    id: PropTypes.number.isRequired,
+    name: PropTypes.string.isRequired,
+  }),
+};
+
 class Group extends Component {
```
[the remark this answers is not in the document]
OSPFNeighbour: Done

sdunster:
This is looking much better but I will clean this up and test it works with all the new changes.

sdunster:
I untangled this and rebased it onto master. Just to confirm, there's supposed to be no server changes in this PR, right?
The tip of the old commit graph was 167510f.
If you need to restore it you can use: `git checkout 167510f -b old_pr_97`

sdunster:
Looks good to me with my latest commit but just hang on until I merge it. Re-reviewing the code right now would be a great idea.
screen shot 2017-12-06 at 01 26 50

OSPFNeighbour:
LGTM. fantastic work thats Sam.

OSPFNeighbour:
screenshot 2017-12-07 00 06 26

PR #100, opened 2017-12-03
Upgrade jest

sdunster:
* Beta version because it has a bug fix for testing rejected promises:
  * https://github.com/facebook/jest/pull/4884
* Needed for upcoming unit tests for events/schedules API


sdunster:
@cmiceli  FYI the whole of react native is beta

PR #102, opened 2017-12-04
Rollback to slightly older jest beta

sdunster:


sdunster:
Confirmed this worked locally for both:
* existing integration tests that require `nextTick` (broken in newest beta)
* new unit tests that require fixed promise rejection code (broken in `latest`)

I'll wait for travis to confirm everything is OK there but I don't really trust it since it didn't notice the breakage last time...

PR #105, opened 2017-12-04
Handle users with invalid auth tokens more gracefully

sdunster:
* If a user has a valid JWT but it points to an invalid user or device we return an unauthorized message which gets caught by the client and automatically logs them out with a cute error message.
* This also moves the logic we have that fakes a login as our test user for graphiql and other requests with no auth header.


sdunster:
Based on #104

PR #107, opened 2017-12-04
Properly wipe the apollo store on logout

### LINKED ISSUE #15: See someone else's details in settings
Repro:
* sign up/create account (i used sdunster / my email / test)
* tap settings (see: ttdykes/ his email)

sdunster:
* This was a long standing bug that would manifest as you appearing to see another user's details when logging in and out. It would basically show you cached data for the previously logged in user.
* This fixes #15 finally.


sdunster:
Verified login and signup works and the information shown in settings, groups and schedules is what I expect each time.

PR #110, opened 2017-12-04
Prevent duplicate group join server side

### LINKED ISSUE #109: Ensure the server join group mutation does not allow a user to join a group more than once

OSPFNeighbour:
fixed #109 

Server side code to check membership first

test to verify new function

new test function to look for a failure

Collapsed group member tests into single group

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -311,7 +311,13 @@ export const groupHandler = {
         if (!group) {
           return Promise.reject(Error('Invalid group!'));
         }
-        return group.addUser(user).then(() => group);
+        return user.getGroups({ where: { id: groupId } }).then((existing) => {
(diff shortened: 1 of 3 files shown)


server/src/logic.js
```diff
@@ -311,7 +311,13 @@ export const groupHandler = {
         if (!group) {
           return Promise.reject(Error('Invalid group!'));
         }
-        return group.addUser(user).then(() => group);
+        return user.getGroups({ where: { id: groupId } }).then((existing) => {
+          console.log(existing);
```
sdunster: remove

OSPFNeighbour: aaaw. i had even fixed that and forgot to save before pushing

PR #117, opened 2017-12-07
TimeSegment Resolvers & Mutations + Schedule start/stop

OSPFNeighbour:
Mutations/log/schema for

editTimeSegment - edit existing segment
removeTimeSegment - remove existing segment
createTimeSegment - create new segment

Added start/end time to schedules to support perpetual vs short term schedules.

**NO JEST TESTS AT THIS STAGE DUE TO ASYNC ISSUES, HAVE TESTED EACH QUERY LOCALLY**

### DIFF
--- a/server/src/constants.js
+++ b/server/src/constants.js
@@ -1,0 +1,3 @@
+const distantFuture = 2147483647;
+
+export default distantFuture;
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -13,7 +13,7 @@ const creators = {
     if (!group || !group.id) {
       return Promise.reject(Error('Must pass group'));
     }
-    if (!startTime || !endTime) {
+    if (typeof startTime === 'undefined' || typeof endTime === 'undefined') {
       return Promise.reject(Error('Must pass start and end times'));
     }
     return Schedule.create({
@@ -121,7 +121,7 @@ const creators = {
     if (!schedule || !schedule.id) {
       return Promise.reject(Error('Must pass schedule'));
     }
-    if (!startTime || !endTime) {
+    if (typeof startTime === 'undefined' || typeof endTime === 'undefined') {
       return Promise.reject(Error('Must pass start and end times'));
     }
     return TimeSegment.create({
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -13,6 +13,9 @@ const creators = {
     if (!group || !group.id) {
       return Promise.reject(Error('Must pass group'));
     }
+    if (!startTime || !endTime) {
+      return Promise.reject(Error('Must pass start and end times'));
+    }
     return Schedule.create({
       name,
       details,
@@ -117,6 +120,9 @@ const creators = {
     }
     if (!schedule || !schedule.id) {
       return Promise.reject(Error('Must pass schedule'));
+    }
+    if (!startTime || !endTime) {
+      return Promise.reject(Error('Must pass start and end times'));
     }
     return TimeSegment.create({
       status,
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -9,13 +9,15 @@ const creators = {
   organisation: ({ name }) =>
     Organisation.create({ name }),
 
-  schedule: ({ name, details, group }) => {
+  schedule: ({ name, details, startTime, endTime, group }) => {
     if (!group || !group.id) {
       return Promise.reject(Error('Must pass group'));
     }
     return Schedule.create({
       name,
       details,
+      startTime,
+      endTime,
       groupId: group.id,
     });
   },
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -268,7 +268,7 @@ export const timeSegmentHandler = {
         }),
       );
   },
-  editTimeSegment(_, args, ctx) {
+  updateTimeSegment(_, args, ctx) {
     const { segmentId, status, startTime, endTime } = args.timeSegment;
     return getAuthenticatedUser(ctx)
       .then(() =>
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -7,6 +7,7 @@ import {
   Organisation,
   Schedule,
   Event,
+  TimeSegment,
 } from './models';
 import Creators from './creators';
 import { schedulePerms, eventPerms } from './perms';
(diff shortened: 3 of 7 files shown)


server/src/creators.js
```diff
@@ -9,13 +9,15 @@ const creators = {
   organisation: ({ name }) =>
     Organisation.create({ name }),
 
-  schedule: ({ name, details, group }) => {
+  schedule: ({ name, details, startTime, endTime, group }) => {
```
sdunster: enforce the existence of startTime/endTime because (at least for now) a time segment shouldn't exist without them

server/src/logic.js
```diff
@@ -234,6 +235,55 @@ export const timeSegmentHandler = {
...
+            return Promise.reject(Error('Invalid segment!'));
+          }
+          return segment.destroy().then((rows) => {
+            if (rows) { return true; }
+            return false;
+          });
+        }),
+      );
+  },
+  editTimeSegment(_, args, ctx) {
```
sdunster: prefer "update" to "edit" but dw for now

server/src/test-data.js
```diff
@@ -1,5 +1,10 @@
 import Creators from './creators';
 
+const nowInGMT = () => {
+  const d = new Date();
+  return Math.round((d.getTime() - 39600) / 1000);
```
sdunster: you sure this is necessary? I thought that getTime() returned UTC timestamps (unix timestamps should always be in UTC)
either way this is fine for tests but we need to be careful with this stuff in the actual code.

server/src/test-data.js
```diff
@@ -51,11 +56,13 @@ export const loadTestData = () =>
               group,
               name: 'Wollongong S&W',
               details: 'Weekly storm availability',
+              startTime: 0, // ongoing
```
sdunster: ok didn't realise you wanted to support this, sure. You probably want to store an ongoing endTime as some constant timestamp set far into the future though so that DB querying/indexes work.

server/src/test-data.js
```diff
@@ -64,11 +71,15 @@ export const loadTestData = () =>
...
               name: 'Wollongong VR',
               details: 'Weekly VR availability',
+              startTime: 0, // ongoing
+              endTime: 0, // ongoing
             }),
             Creators.schedule({
               group,
               name: 'Emergency Services Expo',
               details: '5 people needed for demonstrations on Sunday',
+              startTime: 1512860400, // Sunday 10/12/17 @ 10am AEST in GMT
```
sdunster: no this isn't the correct stamp to store in the DB.
```
$ TZ=AEST date -d @1512860400
Sat Dec  9 23:00:00 AEST 2017
```

If you want to find the correct stamp use this:
```
$ TZ=AEST date -d '2017-12-10 10:00' +%s
1512900000
```

you can confirm like this:
```
$ TZ=AEST date -d @1512900000
Sun Dec 10 10:00:00 AEST 2017
```

OSPFNeighbour: I dont think your right on that one.

1512860400
Is equivalent to:
12/09/2017 @ 11:00pm (UTC)

Add 11 hrs to that and you get 13/09/2017 @ 10.00am

Both a random handful of website, and the availability app agree with me on this one.

screenshot 2017-12-07 15 10 49

server/src/test-data.js
OSPFNeighbour: Researched... yes you are correct.

server/src/logic.js
OSPFNeighbour: Done

OSPFNeighbour:
@sdunster regarding your far future time stamp for end time, isnt 0 easier because it will never happen, where as any date in the future is a possible value?

server/src/test-data.js
sdunster: Yeah you're totally right - `TZ=AEST` didn't work I guess cause you're in DST.
```
$ TZ=Australia/Sydney date -d @1512860400
Sun Dec 10 10:00:00 AEDT 2017
```
Your latest comment said this:
> Add 11 hrs to that and you get 13/09/2017 @ 10.00am

But I think you mean:
> Add 11 hrs to that and you get 12/10/2017 @ 10.00am

unrelated: why are you using an american locale?

sdunster:
When I say far future I mean like INT_MAX far. If we're storing i64s in the DB then that is:
> Friday, April 11th 2262, 23:47:16 (GMT)
It seems a little counter intuitive but 0 is just as arbitrary.
The reason why this is useful is we can write a query like this and get ranges of segments (for example if we want to draw a week-long view) and include these without having to specially deal with `0`s:
```
SELECT * FROM timesegments
WHERE endTime > startOfThisWeek
  AND startTime < endOfThisWeek
```
This query would return all segments that overlap with the week we are interested in even if there is no defined start/end date (if we store end date as a far-future date).
Compare to something like this:
```
SELECT * FROM timesegments
WHERE (endTime > startOfThisWeek OR endTime == 0)
  AND (startTime < endOfThisWeek OR startTime == 0)
```
Not only is the query longer and more complex but it's inefficient because the RDBMS needs to dig through an additional part of the index.
This is actually somewhat of a common practice - see https://developer.apple.com/documentation/foundation/nsdate/1415385-distantfuture for example.

It doesn't look like there's anything built into JS but we can just define our own.

server/src/test-data.js
OSPFNeighbour: Blame website for locale.

OSPFNeighbour:
We are only using 32bit Int, so 2147483647, January 19, 2038 03:14:07 GMT

server/src/creators.js
OSPFNeighbour: done.

server/src/constants.js
```diff
@@ -0,0 +1,3 @@
+const distantFuture = 2147483647;
+
+export default distantFuture;
```
sdunster: eh probably don't need to make this default cause we expect to put other things here but don't worry for now

server/src/logic.js
```diff
@@ -234,6 +235,55 @@ export const timeSegmentHandler = {
   user(timesegment) {
     return timesegment.getUser();
   },
+  createTimeSegment(_, args, ctx) {
+    const { scheduleId, status, startTime, endTime } = args.timeSegment;
+    return getAuthenticatedUser(ctx)
+      .then(user =>
```
sdunster: no permissions checks (which is fine for now) just noting it

server/src/resolvers.js
```diff
@@ -129,6 +129,15 @@ export const Resolvers = {
     createSchedule(_, args, ctx) {
       return scheduleHandler.createSchedule(_, args, ctx);
     },
+    createTimeSegment(_, args, ctx) {
+      return timeSegmentHandler.createTimeSegment(_, args, ctx);
```
sdunster: `_` is supposed to be unused. If the function you're calling doesn't need it don't pass it. If you want to pass it then name it something other than `_`

sdunster: I should have caught this in previous reviews...

OSPFNeighbour: I just copied the code above, should we clean them all up?

PR #119, opened 2017-12-09
Events screen uses actual data from API

sdunster:
* Query for event details via event-by-ID API
* Display responses
* Support pull to refresh

Caveats:

* Map still static for now
* Test data has more that one entry for a single user which causes react native warnings that will go away.

### DIFF
--- a/server/src/test-data.js
+++ b/server/src/test-data.js
@@ -99,14 +99,32 @@ export const loadTestData = () =>
               details: 'VR operators required to assist with rescue',
               group,
             }).then(event =>
-              Creators.eventResponse({
-                status: 'On Route',
-                detail: 'Attending',
-                destination: 'Scene',
-                eta: 1517443200,
-                event,
-                user,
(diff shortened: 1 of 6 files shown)


server/src/test-data.js
```diff
@@ -86,15 +86,32 @@ export const loadTestData = () =>
...
+                  detail: 'ETA 10 min',
+                  destination: 'Scene',
+                  eta: 1517443200,
+                  event,
+                  user,
+                }),
+                Creators.eventResponse({
+                  status: 'Responding',
+                  detail: 'WOL39',
+                  destination: 'Scene',
```
OSPFNeighbour: down the road I think we should have defined somewhere the options for destination. might be that groups have meeting places, or organisations have meeting places.

sdunster: agreed

OSPFNeighbour:
Fix your lint errors :p

PR #126, opened 2017-12-12
Tdykes.fix.group.refresh

### LINKED ISSUE #114: Pull-to-refresh causes a crash on group detail page
Repro on iOS - unsure if Android affected:
Go to groups tab. Tap a group (tested with Wollongong group). Pull to refresh results in app crash and I get kicked back to springboard.

OSPFNeighbour:
Fixes #114

bind error on function

### DIFF
--- a/client/src/screens/group.screen.js
+++ b/client/src/screens/group.screen.js
@@ -182,6 +182,7 @@ class Group extends Component {
     super(props);
     this.joinGroup = this.joinGroup.bind(this);
     this.leaveGroup = this.leaveGroup.bind(this);
+    this.onRefresh = this.onRefresh.bind(this);
   }
 
   onRefresh() {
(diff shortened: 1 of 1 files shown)


client/src/screens/group.screen.js
```diff
@@ -206,8 +207,6 @@ class Group extends Component {
 
   leaveGroup() {
     this.props.leaveGroupQry({ groupId: this.getGroup().id }).then(() => {
-      // pull group details down again to populate the group details
-      this.onRefresh();
```
sdunster: this seems like an unrelated change?

OSPFNeighbour: It's semi related as it's a totally useless refresh that was working while the other failed.

sdunster: 👍

PR #148, opened 2018-01-08
Deploy master branch automatically to heroku

sdunster:
* Use travis build stages to ensure the tests pass before we push.
* New build.sh script to transpile and bundle.
* We ship only the transpiled source, `package.json`, `yarn.lock` and `Procfile` to heroku which should allow for faster dyno start times.
* No longer doing weird git subtree stuff - just upload tarball direct from travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,20 @@ env:
   - COMPONENT=server MODE=lint
   - COMPONENT=server MODE=test-integration
   - COMPONENT=server MODE=test-unit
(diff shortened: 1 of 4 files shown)


sdunster:
Tested with `heroku` branch. That branch is running on heroku now - proof that it works: https://ses-availability-api.herokuapp.com/graphiql
When this lands it should deploy from master. We won't know for sure if that works until it lands.

PR #149, opened 2018-01-11
search schedule by ID

OSPFNeighbour:
allows you to pull a single schedule out by passing ID

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -72,7 +72,10 @@ export const userHandler = {
         where: { groupId: groups.map(g => g.id) },
       }));
   },
-  schedules(user) {
+  schedules(user, args) {
+    if (args.id) {
+      return Schedule.findAll({ where: { id: args.id } });
+    }
     return user.getGroups()
       .then(groups => Schedule.findAll({
         where: { groupId: groups.map(g => g.id) },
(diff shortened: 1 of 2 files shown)


server/src/logic.js
```diff
@@ -72,7 +72,10 @@ export const userHandler = {
         where: { groupId: groups.map(g => g.id) },
       }));
   },
-  schedules(user) {
+  schedules(user, args) {
+    if (args.id) {
+      return Schedule.findAll({ where: { id: args.id } });
```
sdunster: this should only ever return schedules that the user is part of because this is exposed from a user. If you have a valid reason to filter so that it returns a user's schedule by ID then you should add `id` to the where clause below on L78. Otherwise (preferred) just make your client use the top level query which will go through the logic at `scheduleHandler.query` on L199.

server/src/schema.js
```diff
@@ -95,7 +95,7 @@ export const Schema = [`
     username: String # this is the name we'll show other users
     organisation: Organisation! # the organisation this user belongs to
     groups(id: Int): [Group]! # groups the user belongs to
-    schedules: [Schedule]! # schedules from groups the user belongs to
+    schedules(id: Int): [Schedule]! # schedules from groups the user belongs to
```
sdunster: we already have this as a top level query on L173.

sdunster:
Closing this because it shouldn't be necessary.

PR #150, opened 2018-01-14
Remove all function binds

sdunster:
* Switch to using new syntax where it is not required. Thanks @sjmcculloch!
* Remove most constructors as they are no longer necessary.


OSPFNeighbour:
so the binding in the constructor just happen automagically?

sdunster:
Yeah if you use the following syntax to define functions it binds for you:
```
class Foo {
  bar = () => {
    // this is bound automatically - use this when the function is used as a callback
  }
}
```

Versus this which does not
```
class Foo {
  bar() {
    // this is not bound automatically. Use this when not used as a callback.
  }
}
```

PR #151, opened 2018-01-14
Fix PropTypes validation failing

sdunster:
* start and end time of a schedule could be undefined and therefore should not be `isRequired`.

### DIFF
--- a/client/src/screens/schedules.screen.js
+++ b/client/src/screens/schedules.screen.js
@@ -103,8 +103,8 @@ Schedule.propTypes = {
     id: PropTypes.number,
     name: PropTypes.string,
     details: PropTypes.string,
-    startTime: PropTypes.number.isRequired,
-    endTime: PropTypes.number.isRequired,
+    startTime: PropTypes.number,
+    endTime: PropTypes.number,
   }),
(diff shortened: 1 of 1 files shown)


OSPFNeighbour:
errr. but shouldnt be not allow it to be undefined? whats the use case for allowing undefined?

sdunster:
The way I see it there is 2 types of schedules:
* Schedule defined for a particular time range ie this weekend for storm jobs or something
* Schedule with no defined date range because it is ongoing forever. It might still have a start time which is the creation time, I guess.
Either way I think it's a change that makes it easier for us to test right now and we can change back later if it doesn't fit the model.

OSPFNeighbour:
We had a big discussion about this previously which lead to the creation of the distantFuture constant as you rightfully pointed out you cant sort or query easily for undefined, or 0 in the database.

Perpetuals have been defined as ending distantFuture and starting at 0 or 'now'

sdunster:
Oh crap I totally forgot about this. Good call. I will make it required and modify the creator to enforce it is set.

sdunster:
will deal with this later

PR #155, opened 2018-01-22
New test data folder and bulk user creation

OSPFNeighbour:
Its not really how I want it to look in the long run, but its serving the purpose I needed which is more test users in groups with events and schedules.


OSPFNeighbour:
@sdunster if you could help me with the test coverage for the promise.all inside my code that would be appreciated.

server/src/__tests__/constants.js
```diff
@@ -1,5 +1,14 @@
...
+  expect(distantFuture).toBeDefined();
+  expect(nowInUTC).toBeDefined();
 });
+
+it('nowInUTC returns a number', () =>
+  expect(nowInUTC()).toBeGreaterThan(0),
+);
+
+it('distantFuture returns a number', () =>
+  expect(distantFuture).toEqual(2147483647),
```
sdunster: better:
```
expect(distantFuture).toBeGreaterThan(new Date().convertToUnixTime()),
```

server/src/constants.js
```diff
@@ -1,3 +1,6 @@
-const distantFuture = 2147483647;
+export const distantFuture = 2147483647;
 
-export default distantFuture;
+export const nowInUTC = () => {
```
sdunster: this really doesn't belong in `constants.js` you could make this `time.js` or `util.js`...

server/src/__tests__/constants.js
sdunster: (make sure it's converted into seconds not milliseconds otherwise the test won't be valid)
maybe add 100 years to the timestamp too :P

PR #160, opened 2018-01-30
Availability fixes

sjmcculloch:
- Fixed hidden lint errors 
- Changed hardcoded styles in components to use /themes/colors
- Removed unused TextInput,  Header, Number components
- Moved status bar to container (away from screens)


client/src/components/Container/Center.js
```diff
@@ -6,8 +6,7 @@ import styles from './styles';
 const Center = ({ children }) => <View style={styles.containerCenter}>{children}</View>;
 
 Center.propTypes = {
-  // eslint-disable-next-line
-  children: PropTypes.any,
+  children: PropTypes.node.isRequired,
```
sdunster: 👍

client/src/components/Container/Container.js
```diff
@@ -10,12 +10,16 @@ const Container = ({ children, isAlt }) => {
     containerStyles = [styles.containerAlt];
   }
 
-  return <View style={containerStyles}>{children}</View>;
+  return (
+    <View style={containerStyles}>
+      <StatusBar barStyle="light-content" />
```
sdunster: hang on isn't `Container` used like all over the place? Unless I misunderstand `StatusBar` should probably only ever exist once at any point in time in the view hierarchy. I would have added it to the root application node in `app.js`. Maybe take a look at how some other react native apps do it?

OSPFNeighbour: It is possible to have multiple StatusBar components mounted at the same time. The props will be merged in the order the StatusBar components were mounted.

sjmcculloch: We can just remove - and address later - it was just for styling.

PR #163, opened 2018-01-31
fixed broken leave group code so that you can leave a group.

OSPFNeighbour:
Fixed bad binding on leaveGroup function.

Changed leaveGroup GQL code to take advantage of query cache which involved having the server return the group after leaving. TODO; preemptive cache the query

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -382,7 +382,7 @@ export const getHandlers = ({ models, creators: Creators }) => {
               return Promise.reject(Error('Invalid group!'));
             }
             return group.removeUser(user).then((rows) => {
-              if (rows) { return true; }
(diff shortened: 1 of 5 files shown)


server/src/logic.js
```diff
@@ -382,7 +382,7 @@ export const getHandlers = ({ models, creators: Creators }) => {
               return Promise.reject(Error('Invalid group!'));
             }
             return group.removeUser(user).then((rows) => {
-              if (rows) { return true; }
+              if (rows) { return Group.findById(groupId); }
```
sdunster: couldn't you just return the `group`. I assume `removeUser()` updates the internal state automatically. Maybe not but worth a try since it would save a DB query.
```
if (rows) { return group; }
```

PR #164, opened 2018-02-01
return only schedules that have time segments.

OSPFNeighbour:
amirite?

screenshot 2018-02-01 13 00 33
screenshot 2018-02-01 13 00 19

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -1,5 +1,7 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
+import Sequelize from 'sequelize';
+
 import JWT_SECRET from './config';
 import { schedulePerms, eventPerms } from './perms';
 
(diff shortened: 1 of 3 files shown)


sdunster:
@sjmcculloch this is on you - you're working on this stuff. Review, or take what you need and close it.

server/src/logic.js
```diff
@@ -343,7 +345,17 @@ export const getHandlers = ({ models, creators: Creators }) => {
       users(group) {
         return group.getUsers();
       },
-      schedules(group) {
+      schedules(group, args) {
+        const { hasTimeSegments } = args;
+        console.log('MARK', hasTimeSegments);
```
sjmcculloch: Can we remove this console.log

OSPFNeighbour:
just confirming we still want this to be merged?

sjmcculloch:
It's not strictly needed at the moment (the code does the filtering) - we will move this to a separate query eventually so either way is OK.

sdunster:
its cool we can just leave this around for now.

PR #168, opened 2018-02-01
quick and nasty code to draw a few markers on the map. 

OSPFNeighbour:
Code stolen from react-native-maps example folder.

Markers are static and drawn from mock data.

screenshot 2018-02-01 22 06 35


sdunster:
ios
screen shot 2018-02-01 at 17 30 42

PR #171, opened 2018-02-02
Mock event response screen

sdunster:
* Icons on response buttons
* WIP: styles, components


sdunster:
screen shot 2018-02-01 at 18 47 54

PR #172, opened 2018-02-05
test-data overhaul

sdunster:
* Support creation of a wide range of test data from simple JSON
* Needed for testing event stuff
* Better data for showing off the app/demos
* Split test data creation into small functions
* Move under `test-data/` directory with `fixtures.js` with raw data and `load.js` for actual code.
* Creates everything the previous versin did _except_ `Schedule` `TimeSegments` - but this should be easy to add.


OSPFNeighbour:
I would suggest making your default group 'NSW SES' or something state level, and a secondary for Kiama/Parramatta.

State want to use this to collect availability from 'everyone' so having everyone default to membership in a large group would be good.

sdunster:
I see what you mean. The default was just intended to be for while we are testing. Feel free to make the change if you'd like - just make sure the `test` user is in some interesting events so I can test easily.

PR #179, opened 2018-02-07
DisplayName change on settings page

OSPFNeighbour:
UI for DisplayName change
mutations to support future profile changes

+ Bonus empty modal component

screenshot 2018-02-07 15 37 13

screenshot 2018-02-07 15 35 45

screenshot 2018-02-07 15 35 51

### DIFF
--- a/client/src/components/Modal/Modal.js
+++ b/client/src/components/Modal/Modal.js
@@ -1,0 +1,25 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { View, Button, Modal } from 'react-native';
+
+import styles from './styles';
+
+const BasicModal = ({ visible, closeModal, children }) => (
+  <Modal visible={visible} animationType="slide" onRequestClose={closeModal}>
+    <View style={styles.modalContainer}>
+      <View style={styles.innerContainer}>
+        {children}
+        <Button onPress={closeModal} title="Close modal" />
+      </View>
+    </View>
+  </Modal>
+);
+
+BasicModal.propTypes = {
+  children: PropTypes.node,
+  closeModal: PropTypes.func.isRequired,
+  visible: PropTypes.bool.isRequired,
+};
+
+export default BasicModal;
--- a/client/src/screens/settings.screen.js
+++ b/client/src/screens/settings.screen.js
@@ -4,11 +4,14 @@ import React, { Component } from 'react';
 import { ActivityIndicator, Alert, Button, Text, View } from 'react-native';
 import { connect } from 'react-redux';
 import { graphql, compose } from 'react-apollo';
-import Icon from 'react-native-vector-icons/FontAwesome';
+import Prompt from 'react-native-prompt';
 
+import Icon from 'react-native-vector-icons/FontAwesome';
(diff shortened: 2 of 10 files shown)


client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,25 @@
...
+  </Modal>
+);
+
+BasicModal.propTypes = {
+  children: PropTypes.node,
+  closeModal: PropTypes.func.isRequired,
+  visible: PropTypes.bool.isRequired,
+};
+
+export default BasicModal;
```
sdunster: if this module is named `Modal` it should probably export something called `Modal` rather than `BasicModal` right? @sjmcculloch

client/src/screens/settings.screen.js
```diff
@@ -101,6 +113,10 @@ class Settings extends Component {
     tabBarIcon: ({ tintColor }) => <Icon size={28} name="cog" color={tintColor} />,
   };
 
+state = {
```
sdunster: indendation. how come lint didn't find this? :/

client/src/screens/settings.screen.js
```diff
@@ -137,48 +153,79 @@ class Settings extends Component {
...
+    this.setState({ promptVisible: false });
+    this.props
+      .updateUserProfile({ displayName: value });
+  }
+
+  openPrompt = () => {
+    this.setState({ promptVisible: true });
+  };
+
+    closePrompt = () => {
```
sdunster: is it necessary to have separate functions for close vs cancel?

client/src/screens/settings.screen.js
```diff
@@ -201,4 +248,7 @@ const mapStateToProps = ({ auth }) => ({
   auth,
 });
 
-export default compose(connect(mapStateToProps), userQuery, updateLocationMutation)(Settings);
+export default compose(connect(mapStateToProps),
```
sdunster: suggest whitespace like this:
```
export default compose(
  connect(mapStateToProps),
  userQuery,
  updateUserProfileMutation,
  updateLocationMutation,
)(Settings);
```

client/src/components/Modal/Modal.js
OSPFNeighbour: cant without getting a "name already in use" lint warning

client/src/screens/settings.screen.js
OSPFNeighbour: in future, maybe? I can collapse if we dont care

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { View, Button, Modal } from 'react-native';
```
sdunster: change this to
```
import { View, Button, Modal as RNModal } from 'react-native';
```

client/src/components/Modal/Modal.js
sdunster: see above suggestion

client/src/screens/settings.screen.js
sdunster: at least fix indentation plz

PR #181, opened 2018-02-08
Start of hook code

cmiceli:
This is a diff to begin the hook code for the ingestion of events from other services into availability. The factory allows us to add other functions over time, so other use cases are easily merged in.

### DIFF
--- a/server/src/callback.js
+++ b/server/src/callback.js
@@ -1,0 +1,22 @@
+function eventCallback(req, res) {
+  // check auth
+  if (req.header('Bearer')) {
+    console.log(req.header('Bearer'));
+    console.log(req.header('Authorization'));
+  }
+
+  console.log(req.body);
+
+  // get the base handler to add the event
+  // send notifications (probably belongs in the other code)
+  res.send('OK');
+}
+
+function getCallback(type) {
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
Fix the lint issues, obviously.

server/src/callback.js
```diff
@@ -0,0 +1,18 @@
...
+    console.log(req.header('Bearer'));
+    console.log(req.header('Authorization'));
+  }
+
+  // get the base handler to add the event
+  // send notifications (probably belongs in the other code)
+  res.send('OK');
+}
+
+export default function getCallback(type) {
```
sdunster: i dont really understand the need for this. Couldn't you just `app.use()` and pass the `eventCallback` function directly?

server/src/callback.js
```diff
@@ -0,0 +1,18 @@
+function eventCallback(req, res) {
+  // check auth
+  if (req.header('Bearer')) {
+    console.log(req.header('Bearer'));
+    console.log(req.header('Authorization'));
```
sdunster: cool but I don't think we really need auth to begin with.

Issue #95, opened 2017-12-03
Android native navigation buttons don't work


Sprooty:
The native android launcher buttons for navigation don't appear to interact with the app - pushing the back arrow closed the application. On other apps, the back button will feed back into the app and take you to  the previous screen and or out of a menu within the app.

sdunster:
ack - this is something we'll need to look at. Atm we're using modal navigation style. This might be handled automatically if we switch to using stack based style (which is what we should be using for everything except login/signup probably anyway)

sdunster:
tim is working on this too maybe

OSPFNeighbour:
Need to use Backhandler react element. It looks easy I just don't understand where to slot it in.

OSPFNeighbour:
FYI - not as easy as i thought. I give up and will leave this for @sdunster or @sjmcculloch

PR #182, opened 2018-02-08
hacky implementation of BackHandler to make the android users happy

OSPFNeighbour:
addresses #95


OSPFNeighbour:
The user experience is almost perfect, if I should be hooking it in better
please point me in the right direction

PR #184, opened 2018-02-08
Interface for editing an event response

sdunster:
* UI is primitive. Displays detail and allows you to press a button to change your status.
* Updating the cache may or may not work.
* Actual updating the server is NYI - separate diff and then re-enable on client side.

### DIFF
--- a/client/src/screens/event-detail.screen.js
+++ b/client/src/screens/event-detail.screen.js
@@ -4,6 +4,7 @@ import {
   FlatList,
   ActivityIndicator,
   Text,
+  Button,
   View,
   Dimensions,
 } from 'react-native';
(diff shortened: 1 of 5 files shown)


sdunster:
screen shot 2018-02-08 at 00 06 42
screen shot 2018-02-08 at 00 06 37

client/src/screens/event-detail.screen.js
```diff
@@ -203,12 +224,15 @@ class EventDetail extends Component {
...
         response={item[1]}
+        auth={this.props.auth}
+        onEdit={this.editResponse}
       />
     );
   };
 
   render() {
     const { event, loading, networkStatus } = this.props;
+    console.log(event);
```
OSPFNeighbour: ?

PR #185, opened 2018-02-08
Rudimentary 5s refresh for event details

sdunster:
based off of `sdunster.event_response`


sdunster:
this isn't exactly visible currently but I confirmed it works by looking at server logs

PR #191, opened 2018-02-09
Smcculloch.navigation deep links

sjmcculloch:
- The deep links on the home screens for events & requests will now reset the stack before navigating (via a redux call)

- Moved navigation reducers into state folder (note: standard is to have these in /reducers /actions /store - we should move them)


sdunster:
yeah nice feel free to move stuff around just do it in a separate PR to adding features etc

PR #195, opened 2018-02-12
Selector & Client Tests

sjmcculloch:
- Removed existing tests in client/__tests__
- Configured Jest to look in __tests__ for anything *.test.js (to avoid conflict with fixtures)
- Moved filtering logic for timeSegments from availability index component into /selectors/schedule.js
- Added tests for reducers, actions & selectors


client/src/screens/availability/Index.js
```diff
@@ -86,43 +86,22 @@ class Index extends Component {
...
-            endTime: timeSegment.endTime,
-            id: timeSegment.id,
-            status: timeSegment.status,
-          });
-        }
-      });
-    });
-
-    filteredItems.sort((a, b) => (a.startTime > b.startTime ? 1 : -1));
+    const filteredItems = selectSchedules(user.schedules, { startTime, endTime });
```
sdunster: 👍

client/src/selectors/schedules.js
```diff
@@ -0,0 +1,24 @@
+export default (schedules, { startTime, endTime }) => {
+  const filteredItems = [];
+  schedules.map(schedule =>
+    schedule.timeSegments.map((timeSegment) => {
+      if (
+        (timeSegment.startTime === 0 || timeSegment.startTime >= startTime) &&
+        (timeSegment.endTime === 2147483647 || timeSegment.endTime <= endTime)
```
sdunster: please use the constants in place of `0` and `2147483647`. I'm also pretty sure you shouldn't need to special-case these but I can't think of the logic right now.

PR #196, opened 2018-02-13
Tdykes.ui.map.markers

OSPFNeighbour:
Am I on the right track with this UI work?


client/src/screens/event-detail.screen.js
```diff
@@ -12,10 +12,14 @@ import MapView, { Marker } from 'react-native-maps';
...
 import Icon from 'react-native-vector-icons/FontAwesome';
 import { connect } from 'react-redux';
+import moment from 'moment';
 import _ from 'lodash';
 
 import { extendAppStyleSheet } from './style-sheet';
 import EVENT_QUERY from '../graphql/event.query';
+import markers from '../assets/images/map/markers';
+
+import { UserMarker } from '../components/MapMarker/';
```
sdunster: drop trailing slash

sdunster:
rebase

client/src/assets/images/map/markers/index.js
```diff
@@ -0,0 +1,18 @@
+import crosshairs from './crosshairs.png';
+import lhq from './lhq.png';
+import pin from './pin.png';
+
+const mapMarker = (name) => {
```
sjmcculloch: Why is this component in assets/images? Could it not be a stateless component in /components ?

client/src/components/MapMarker/UserMarker.js
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <View style={styles.container}>
+      <View style={styles.bubble}>
+        <Text style={[styles.name]}>{name}</Text>
+      </View>
+      <View style={styles.arrowBorder} />
+      <View style={styles.arrow} />
+    </View>
+  );
+};
```
sjmcculloch: Fyi - you can convert this to:
```
const UserMarker = ({ name }) => (
    <View style={styles.container}>
      <View style={styles.bubble}>
        <Text style={[styles.name]}>{name}</Text>
      </View>
      <View style={styles.arrowBorder} />
      <View style={styles.arrow} />
    </View>
  );
```

client/src/components/MapMarker/styles.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  name: {
+    color: '#FFFFFF',
+    fontSize: 13,
+  },
+  arrow: {
+    backgroundColor: 'transparent',
+    borderWidth: 4,
+    borderColor: 'transparent',
+    borderTopColor: '#FF5A5F',
```
sjmcculloch: You can avoid hardcoding colors by referencing:

`import Colors from '../../themes/Colors';`


 `backgroundColor: Colors.bgBtn,`

client/src/assets/images/map/markers/index.js
OSPFNeighbour: I wasnt sure on the normal for something like this. react wont let you dynamic load images, so this code is kinda an asset wrapper so I thought it could like with the assets it loads?

client/src/components/MapMarker/UserMarker.js
OSPFNeighbour: sweet!

client/src/components/MapMarker/styles.js
OSPFNeighbour: in future we will need a few different colors for this, ill hold off on linking back to theme colors until i link that code in

OSPFNeighbour:
rebased against master, squashed a few commits down. I should get a PhD for that effort.

client/src/assets/images/map/markers/index.js
sjmcculloch: Should be OK - we may revisit - I typically keep all the stateless components together - let's proceed so we can keep things moving.

client/src/components/MapMarker/styles.js
sjmcculloch: ok - sounds great - typically we would have a few set theme colors we re-use - then we can tweak in one place if changes are required.

PR #197, opened 2018-02-13
Tdykes.response.mapmarkers.realtionship

OSPFNeighbour:
destination on event response now points to the event location directly.

Updated UI to work, but in next rework of the events screen it will take advantage of this, currently it just doesnt do anything special with it, I just made it not break

land #201 first

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -336,15 +336,23 @@ export const getHandlers = ({ models, creators: Creators }) => {
             const updateArgs = { ...args.response };
             delete updateArgs.id;
 
-            if (result.length > 0) {
-              return result[0].update(updateArgs);
-            }
+            return event.getEventlocations({ where: { id: updateArgs.destination.id } })
+              .then((destination) => {
+                if (!destination) {
+                  return Promise.reject(Error('Unknown destination passed'));
+                }
+                updateArgs.eventlocationId = (destination ? destination.id : null);
 
-            return Creators.eventResponse({
-              event,
-              user,
-              ...updateArgs,
-            });
+                if (result.length > 0) {
+                  return result[0].update(updateArgs);
+                }
+
+                return Creators.eventResponse({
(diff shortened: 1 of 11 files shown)


OSPFNeighbour:
This is GTG if I could get some stamps. @sdunster 

```
{
                    "user": {
                      "username": "kiama1"
                    },
                    "destination": {
                      "id": 1,
                      "name": "lhq",
                      "detail": "Local Head Quarters"
                    }
                  }
```

server/src/logic.js
```diff
@@ -328,6 +328,7 @@ export const getHandlers = ({ models, creators: Creators }) => {
         const {
           id,
         } = args.response;
+        console.log(args);
```
sjmcculloch: remove consol.log?

sjmcculloch:
It looks OK to me - I'll leave to @sdunster  to stamp for approval.

server/src/logic.js
OSPFNeighbour: thx. removed.

OSPFNeighbour:
land #201 first

server/src/logic.js
```diff
@@ -353,6 +361,9 @@ export const getHandlers = ({ models, creators: Creators }) => {
       user(response) {
         return response.getUser();
       },
+      eventLocations(response) {
```
sdunster: i think you should name this `destination` to match the schema. at the very least it should not be plural

server/src/resolvers.js
```diff
@@ -79,6 +79,9 @@ export const getResolvers = (handlers) => {
       user(response, args, ctx) {
         return eventResponseHandler.user(response, args, ctx);
       },
+      destination(response, args, ctx) {
+        return eventResponseHandler.eventLocations(response, args, ctx);
```
sdunster: change this to `destination` as well

PR #198, opened 2018-02-13
added fields sourceIdentifier, permalink to events

OSPFNeighbour:
```
"events": [
              {
                "name": "VR at Gerringong Falls",
                "details": "2 climbers stranded 50m from top of cliff.",
                "location": "-34.66828, 150.85219",
                "sourceIdentifier": "654-321",
                "permalink": "https://jobssytem.com/jobs/c33367701511b4f6020ec61ded352059"
              },
              {
                "name": "FR - WEBB STREET, NORTH PARRAMATTA NSW 2151",
                "details": "APPROX 2 VEH SUBMERGED IN STREET AA WHICH IS CURRENTLY FLOODED JUST OVER WHEEL HEIGHT ON A STANDARD VEH. UNKNOWN IF ANY PERSONS OB. TREE DOWN AND NO WIRES DOWN. NFI",
                "sourceIdentifier": "111-222",
                "permalink": "https://jobssytem.com/jobs/111-222"
              }
            ]
```

### DIFF
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -27,14 +27,13 @@ export const getCreators = (models) => {
       });
     },
 
-    event: ({ name, details, location, sourceIdentifier, permalink, group }) => {
+    event: ({ name, details, sourceIdentifier, permalink, group }) => {
       if (!group || !group.id) {
         return Promise.reject(Error('Must pass group'));
       }
       return Event.create({
         name,
         details,
-        location,
         sourceIdentifier,
         permalink,
         groupId: group.id,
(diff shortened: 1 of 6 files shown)


OSPFNeighbour:
Think of it like pretty address. Will go in the banner and is for human reading, not plotting. Most of the time it would be the same a scene marker but looking at some jobs I can see where we might want it.

Some jobs have GPS pin drops as well as "opposite the petrol station"

sjmcculloch:
It's OK by me. Might be confusing long term around the field "location" - especially if it's only used for description purposes (and the real detail is in event markers). 

I'll leave to Sam to approve.

OSPFNeighbour:
Looking at Beacon it has the same. something from police will have a geo location and free text like "Opposite the pub"

PR #199, opened 2018-02-15
Unify home screen

sdunster:
* This is the first step towards building a notifications-like screen.
* Idea here is to automatically show things that are relevant to the user such as:
  * Ongoing events
  * New schedule requests (request for assistance with specified time frame)
  * Ongoing schedule update request (user is nearing the end of the availability they have filled out)
* As these items are actioned (ie schedule is updated or an event is closed) the items drop out of this list. You could still access a schedule or event from the appropriate tab if you need to get back to it.

### DIFF
--- a/client/src/screens/home/Index.js
+++ b/client/src/screens/home/Index.js
@@ -9,10 +9,52 @@ import { goToEvent, goToRequest } from '../../state/navigation.actions';
 
 import CURRENT_USER_QUERY from '../../graphql/current-user.query';
 
-import { Center, Container, Holder } from '../../components/Container';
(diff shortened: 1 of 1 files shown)


sdunster:
screen shot 2018-02-14 at 16 21 58

sjmcculloch:
You can add spacing to the first item using:

```<Holder margin>
<MyComponent>
</Holder>
```

sdunster:
I decided to leave the margin thing un-addressed. I have a better fix incoming.

sdunster:
See #217

PR #200, opened 2018-02-15
Add start/end time field to events

sdunster:
* To track the time that the event began.
* Events begin with endTime set to `DISTANT_FUTURE` - no end time set.


sdunster:
Rebased

sdunster:
undid `EventMarkers`

PR #207, opened 2018-02-19
Store the event as per https://github.com/NSWSESMembers/availability-poc/wiki/API-(Jobs-Search)

cmiceli:
I'd love to have more of the logic around this to finish that, but I flat out have no idea what we need.

It will require a pub-sub system which I can push to for the downstream notifications, but that is the next step. I think that we can solve that with an in-process approach until we have > 1 app host where we can move to something like redis.

### DIFF
--- a/scripts/hook.py
+++ b/scripts/hook.py
@@ -1,0 +1,137 @@
+import requests
+import json
+
+data = {
+    "Id": 28795,
+    "Identifier": "0002-8795",
+    "Sequence": 0,
+    "FloodAssistanceJob": False,
+    "ReferringAgency": None,
+    "ReferringAgencyReference": None,
+    "SituationOnScene": None,
+    "EvacuationRequired": False,
+    "PeopleInundated": 0,
+    "PeopleExtricated": 0,
+    "PeopleEvacuated": 0,
+    "CreatedOn": "2017-12-11T09:22:00",
+    "CreatedBy": {
+        "Id": 9707,
+        "FirstName": "David",
+        "LastName": "Barrie",
+        "FullName": "David Barrie",
+        "Gender": 1,
+        "RegistrationNumber": "123"
+           },
+    "CallerFirstName": "asdf",
+    "CallerLastName": None,
+    "ICEMSIncidentIdentifier": None,
+    "ContactCalled": True,
+    "CallerPhoneNumber": "1234",
+    "ContactFirstName": None,
+    "ContactLastName": None,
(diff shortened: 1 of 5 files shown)


server/src/callback.js
```diff
@@ -1,20 +1,37 @@
...
+  eventCallback(req, res) {
+    const job = {
+      name: req.body.JobType.Name,
+      identifier: req.body.Identifier,
+      description: req.body.JobType.Description,
+    };
 
-  console.log(req.body);
+    Promise.resolve(this.creator.location({
+      name: req.body.Address.GnafId,
```
OSPFNeighbour: this would be 'scene' IMO. as in, when someone uses the app and says where they are going I see most jobs having "scene" and "lhq" as the default two.

LHQ would be the ords under EntityAssignedTo and could be added as well.

sdunster: 👍

server/src/models.js
```diff
@@ -37,6 +37,7 @@ export const defineModels = (db) => {
 
   const EventModel = db.define('event', {
     name: { type: Sequelize.STRING },
+    external_identifier: { type: Sequelize.STRING },
```
sdunster: plz no underscore

server/src/models.js
```diff
@@ -122,6 +123,7 @@ export const defineModels = (db) => {
   // event responses belong to a combination of user/event
   EventResponseModel.belongsTo(EventModel);
   EventModel.hasMany(EventResponseModel);
+  EventModel.hasOne(EventLocationModel, { as: 'location' });
```
sdunster: maybe call this `primaryLocation` or something to disambiguate it from `locations` which is the collection of all locations relevant to the event ("lhq", "scene" are the 2 we would have most of the time but there might be others like "staging area", "locked gate" we could add the ability to add or something.

scripts/hook.py
```diff
@@ -0,0 +1,137 @@
+import requests
```
sdunster: can you please replace the name and the `400` number throughout this with something else? I realise this was given to you as test data but if this is going to be committed into the history of this project (forever) I'd prefer it was a little more anonymous.

OSPFNeighbour:
Chris can we get this sorted and landed please

cmiceli:
not sure how, but this thing didn't sense me pushing the branch here. Probably due to rebasing half way through, thus requiring the --force on the branch.

PR #212, opened 2018-02-27
installed and linked bugsnag, test button on settings page

OSPFNeighbour:
Ive installed and linked Bugsnag, and applied for an opensource free plan with them for the project (they give free use to opensource projects)

https://www.bugsnag.com

@sdunster @cmiceli @sjmcculloch have I linked this in correctly? never done it before TBH.

Test and seems to work, I get bug reports showing in the bugsnag dashboard

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { AsyncStorage, Alert } from 'react-native';
+import { Client as BugSnagClient } from 'bugsnag-react-native';
 
 import { ApolloProvider } from 'react-apollo';
 import { createStore, applyMiddleware, compose } from 'redux';
@@ -26,6 +27,8 @@ export const firebaseClient = new FirebaseClient({
     return notification.finish(WillPresentNotificationResult.All);
   },
 });
+
+export const bugsnag = (BugSnagClient.typeof ? new BugSnagClient() : undefined);
 
 // middleware for requests
 networkInterface.use([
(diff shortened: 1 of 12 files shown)


client/src/app.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { AsyncStorage, Alert } from 'react-native';
+import { Client } from 'bugsnag-react-native';
```
sdunster: Suggest you name this something less ambiguous:
```
import { Client as BugsnagClient } from 'bugsnag-react-native';
```

client/src/app.js
```diff
@@ -27,6 +28,8 @@ export const firebaseClient = new FirebaseClient({
   },
 });
 
+export const bugsnag = new Client();
```
sdunster: protip: put this in it's own file `bugsnag.js` then import from there so it's separate from the rest of the intermingled garbage in `app.js`

OSPFNeighbour:
It's currently set to submit any unhanded react or core errors. That's it's default. Can be set to submit handled and unhandled from either layer.

I'll test what it does with rebuild & also address your points. Thx.

sjmcculloch:
I'm not an expert in this area around the linking area - haven't had to do it much in the past and I've always stayed bound to expo (no ios/android folders). Has it been extensively tested, does the app still function correctly?

sdunster:
This introduced a bunch of project configuration changes unrelated to this diff for firebase cloud messaging that we ultimately decided to remove. I split the parts we want to keep into 2 branches:

* `sdunster.install_bugsnag` #234 
* `sdunster.configure_bugsnag`

PR #215, opened 2018-03-02
Very quick refactor of the groups pages to get them visually similar

OSPFNeighbour:
moved some shitty buttons into nav bar buttons as well.

Keeping the PR's as small as possible otherwise I will totally fuck everything up with the refactor

screenshot 2018-03-02 13 00 14

### DIFF
--- a/client/src/screens/groups/Index.js
+++ b/client/src/screens/groups/Index.js
@@ -1,21 +1,14 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import {
-  FlatList,
-  ActivityIndicator,
-  Text,
-  TouchableHighlight,
-  View,
-} from 'react-native';
+import { FlatList, ActivityIndicator, Button, Text, TouchableHighlight, View } from 'react-native';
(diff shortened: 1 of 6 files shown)


client/src/screens/groups/Index.js
```diff
@@ -116,11 +118,17 @@ Group.propTypes = {
   }),
 };
 
-class Groups extends Component {
-  static navigationOptions = {
+class Index extends Component {
```
sdunster: pretty sure the idea of Index is supposed to be a default for this directory or whatever. I don't really agree with naming a component `Index`. cc @sjmcculloch

sdunster: Also this name isn't actually used outside of this file because you export default the decorated result at the bottom

client/src/screens/groups/StackGroups.js
```diff
@@ -0,0 +1,30 @@
+import { StackNavigator } from 'react-navigation';
```
sdunster: nice work separating this out.

PR #217, opened 2018-03-03
Make <ListItem> use marginTop instead of marginBottom

sdunster:
* There are currently no other users of this component.
* For the sole current use-case (#199) it is better to put the margin at the top rather than the bottom.
* If we need to be able to support using marginBottom again this can be added in as a flag/option.
* Another potential future improvement would be to detect when a cell is the last item and have it add marginBottom as well.


OSPFNeighbour:
Its now pissing me off in #220 as well.

PR #221, opened 2018-03-08
couldn't set a new event response via the mutation

OSPFNeighbour:
an object that should have existed was not getting passed to the creator.


OSPFNeighbour:
Not that I can find it in the doc... but I think it will only return if theres is a result, so checking for the result for truthy is a valid way to handle this.

sjmcculloch:
Looks good if that truthy statement works.

PR #223, opened 2018-03-11
Remove firebase

sdunster:
* Planning to do native push instead
* Currently it is not set up for iOS and is therefore blocking a new release.
  * Would need to include the entire firebase client API to get it to work.


OSPFNeighbour:
Tested and fixes android builds

PR #225, opened 2018-03-12
removed a console.log and fix an undefined var

OSPFNeighbour:


OSPFNeighbour:
removed a console.log and fix an undefined var

PR #228, opened 2018-03-14
Smcculloch.dashboard updates

sjmcculloch:
- Added People Icons

image

- Added add/edit/remove timeSegment capability

image


server/src/schema.js
```diff
@@ -45,6 +45,7 @@ export const Schema = [
     status: String!
     startTime: Int!
     endTime: Int!
+    userId: Int
```
OSPFNeighbour: I was wondering how long before we needed to do something like this. We are almost at the point of needing some level of user permission or system to allow one user to act on behalf of another.

sjmcculloch:
Is this OK @sdunster ?

OSPFNeighbour:
@sdunster Stamp please

sdunster:
All of the other changes are great - I'm just being protective of the stuff in `server`. Ping me on Discord/Messenger and we can talk some more about this directly.

sjmcculloch:
@sdunster @OSPFNeighbour  - I've updated logic.js so it reflects the latest from master

server/src/logic.js
```diff
@@ -338,23 +340,26 @@ export const getHandlers = ({ models, creators: Creators }) => {
...
-                updateArgs.eventlocationId = destination[0].id;
-                if (result.length > 0) {
-                  return result[0].update(updateArgs);
-                }
-
-                return Creators.eventResponse({
-                  event,
-                  user,
-                  ...updateArgs,
+            if (updateArgs.destination) {
```
sdunster: ok so `destination` becomes optional

server/src/logic.js
```diff
@@ -338,23 +340,26 @@ export const getHandlers = ({ models, creators: Creators }) => {
...
+                  return Creators.eventResponse({
+                    event,
+                    user,
+                    destination: destination[0],
+                    ...updateArgs,
+                  });
                 });
-              });
+            }
+            return result[0].update(updateArgs);
```
sdunster: so if there is no `destination` we bypass the `EventLocations` lookup and always update.
What if there is no `destination` and there is no existing `eventResponse`? If you intentionally didn't want to handle that situation I think we need to check for it.. Keep in mind it also breaks the idempotency of this function though. :( (the idea is you call setResponse() regardless of whether a response exists already and it atomically replaces if one exists or creates a new one if it doesn't)

If you want to add the check and ensure it is clear we don't handle it before L362 add:
```
if (!destination) {
  return Promise.reject(Error('Destination must be provided to respond to a event that does not already have a response'));
}

File: server/src/logic.js [hunk 4]
```diff
@@ -378,6 +383,9 @@ export const getHandlers = ({ models, creators: Creators }) => {
         if (args.id) {
           return org.getGroups({ where: { Id: args.id } });
         }
+        if (args.filter) {
+          return org.getGroups({ where: { name: { [Op.like]: `%${args.filter}%` } } });
```
sdunster: nice. is this safe from injection attacks?

PR #230, opened 2018-03-15
Tweaks to visual and usage of list items

OSPFNeighbour:
screenshot 2018-03-16 00 03 10

### DIFF
--- a/client/src/components/List/ListItem.js
+++ b/client/src/components/List/ListItem.js
@@ -5,26 +5,44 @@ import Icon from 'react-native-vector-icons/FontAwesome';
 
 import styles from './styles';
 
-const ListItem = ({ title, bold, subtitleEllipsis, wide, supertitle, subtitle, icon, onPress }) => (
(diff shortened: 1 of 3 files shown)


client/src/screens/events/Detail.js
```diff
@@ -312,7 +312,7 @@ class EventDetail extends Component {
     return (
       <View style={{ flex: 1 }}>
         <EventHeader event={event} />
-        <View style={{ position: 'absolute', bottom: 0, left: 0, right: 0 }}>
+        <View style={{ position: 'absolute', bottom: 6, left: 0, right: 0 }}>
```
sjmcculloch: eventually it would be good to remove these inline styles

sdunster:
longer term I think we should cap/add ellipsis based on how the text renders rather than number of characters - pretty sure `<Text>` has this functionality built-in.

OSPFNeighbour:
didnt find one built in, there are modules available but seemed over kill. I dont want to elips at all but needed to just to make something workable short term

PR #234, opened 2018-03-22
Install bugsnag

sdunster:
* Link for Android, iOS


sdunster:
Added API keys from #212

PR #246, opened 2018-03-30
Prevent travis from building PRs twice

sdunster:
* It does this because it is building each new branch as well as each PR.
* Lets blacklist all branches except master.
* It should still build all new PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: node_js
 node_js: 6
 cache: yarn
+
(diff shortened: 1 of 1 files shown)


sdunster:
Woohoo. Confirmed - there is only one build on this diff now:
* `continuous-integration/travis-ci/pr`

sdunster:
compare to #247 which has both `pr` and `push`

PR #247, opened 2018-03-30
Travis: automatically deploy new react native bundles

sdunster:
Plan is to have travis run tests, push server and only once the heroku health check passes automatically push new react native/JS bundles to both Android and iOS.


OSPFNeighbour:
do you need land this or is it a test? @sdunster

sdunster:
I intend to land this eventually when it works but WIP for now.

PR #248, opened 2018-03-31
EventResponse logic change

OSPFNeighbour:
Logic is as follows:

Anything that is not passed will be unchanged.
Anything passed will be updated.
Anything null will be deleted/removed (biggest one is destination)

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -340,27 +340,39 @@ export const getHandlers = ({ models, creators: Creators }) => {
           .then((result) => {
             const updateArgs = { ...args.response };
             delete updateArgs.id;
-            if (updateArgs.destination) {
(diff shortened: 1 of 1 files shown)


server/src/logic.js
```diff
@@ -340,27 +340,39 @@ export const getHandlers = ({ models, creators: Creators }) => {
...
-                  delete updateArgs.destination;
-                  if (result.length > 0) {
-                    return result[0].update(updateArgs);
-                  }
-                  return Creators.eventResponse({
-                    event,
-                    user,
-                    destination: destination[0],
-                    ...updateArgs,
+            if (typeof (updateArgs.destination) !== 'undefined') { // null is truthy so we need this
```
sdunster: no it's not:
```
!!null
> false
```

PR #249, opened 2018-03-31
Tdykes.event.response.fixes

OSPFNeighbour:
Tweaked the user subtitle text to be smarter by supporting no ETA (epoc 0)

Needs #248 to land first.

screenshot 2018-03-31 18 07 30

screenshot 2018-03-31 18 08 02

### DIFF
--- a/client/src/screens/events/Detail.js
+++ b/client/src/screens/events/Detail.js
@@ -230,6 +230,12 @@ class EventDetail extends Component {
     });
     const summaryByDestinationUsersString = summaryByDestinationUsersArray.join(', ');
 
+    const userSubtitleArray = [];
+    userSubtitleArray.push(myStatus && myStatus.destination ? `Destination ${myStatus.destination.name.toUpperCase()}` : 'No destination set');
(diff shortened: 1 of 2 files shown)


client/src/screens/events/Detail.js
```diff
@@ -230,6 +230,12 @@ class EventDetail extends Component {
     });
     const summaryByDestinationUsersString = summaryByDestinationUsersArray.join(', ');
 
+    const userSubtitleArray = [];
```
sdunster: this is much better than doing it inline in the JSX

client/src/screens/events/Response.js
```diff
@@ -114,12 +114,13 @@ class EventResponse extends Component {
   };
 
   submitEventResponse = () => {
-    const dst = (this.state.destination !== '' ? { destination: { id: this.state.destination } } : {});
-    const eta = moment().add(this.state.eta, 'minutes').unix();
+    const dst = (this.state.destination !== null ?
```
sdunster: would probably fit on one line like this:
```
const dst = (
  this.state.destination !== null ?  { destination: { id: this.state.destination } } : null
);
```

PR #250, opened 2018-03-31
Tdykes.eventresponse.comments

OSPFNeighbour:
New Modal for generic TextInput via a modal.

Linked eventreponse detail to text input and added it to the flow of Destination -> ETA -> comments.

screenshot 2018-04-01 08 54 27

TODO: Refactor this modal into the number input modal to be more more generic

Prereq is #248 and #249 so land them first

### DIFF
--- a/client/src/components/Modal/TextInputModal.js
+++ b/client/src/components/Modal/TextInputModal.js
@@ -1,0 +1,66 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { View, Modal, TouchableOpacity, TouchableWithoutFeedback, Text, TextInput } from 'react-native';
+
+import styles from './styles';
+
+const TextInputModal = ({ closeModal, backModal, onChangeText, visible, onSave }) => (
+  <View>
+    <Modal
+      transparent
+      visible={visible}
+      onRequestClose={closeModal}
+      animationType="fade"
+    >
+      <TouchableWithoutFeedback
+        onPress={() => closeModal}
+      >
+        <View style={styles.overlayStyle}>
+          <View style={styles.headerContainer}>
+            <TouchableOpacity>
+              <View style={styles.headerStyle}>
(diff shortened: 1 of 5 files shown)


sdunster:
rebase onto the others once they land

client/src/components/Modal/TextInputModal.js
```diff
@@ -0,0 +1,66 @@
...
+const TextInputModal = ({ closeModal, backModal, onChangeText, visible, onSave }) => (
+  <View>
+    <Modal
+      transparent
+      visible={visible}
+      onRequestClose={closeModal}
+      animationType="fade"
+    >
+      <TouchableWithoutFeedback
+        onPress={() => closeModal}
```
sdunster: you should just be able to use the following here. If it doesn't work make sure the method you're ultimately passing in as `closeModal` (from the user of `TextInputModal`) is properly bound.
```
onPress={closeModal}
```

client/src/screens/events/Response.js
```diff
@@ -172,6 +175,31 @@ class EventResponse extends Component {
...
+  handleDetailChange = (answer) => {
+    this.setState({
+      detail: answer !== '' ? answer : null,
+    });
+  }
+
+  handleDetail = () => {
+    this.setState({
+      detailModal: false,
+    }, () => this.submitEventResponse());
```
sdunster: in theory this could be:
```
}, this.submitEventResponse);
```
because `submitEventResponse` has `this` bound correctly because it is using `func = () => {` instead of `func() {`

client/src/components/Modal/TextInputModal.js
OSPFNeighbour: Sweet. fixed.

client/src/screens/events/Response.js
OSPFNeighbour: in theory that seems to work. thx

PR #251, opened 2018-04-03
Travis: skip cleanup for code push deploy step

sdunster:


sdunster:
Tested on `test_travis` branch. The only modification from there is changing `test_travis` back to `master`. Finally we have code push working automatically.  If there is no change to the resulting bundle code push will ignore the new bundle.

https://travis-ci.org/NSWSESMembers/availability-poc/builds/361450986

sdunster:
GOOD LOOKS TO ME

PR #252, opened 2018-04-03
Make web build and cache better, CDN safe

sdunster:
* Remove unneeded static file server and deps
* Configure webpack to build files that are CDN safe (use content hash in
  filename)
  * We should now be able to let the CDN cache files (except `index.html`)
    forever and users will get the new code as soon as `index.html` expires.
  * This saves users data because their user agent can cache the JS forever as
    well.
* Webpack now generates the html entry point from template.html and dynamically
  includes js, css etc
* Per webpack recommendations split the JS into:
  * main (our code)
  * vendor (library code, expected to change less frequently)
* Update some deps
* Move a bunch of deps to dev deps
* Build directly into `web/dist`.
* Merge all files from `web/static` directly into the build root untouched with
  `copy-webpack-plugin`
  * `favicon.png`


sdunster:
Confirmed this works locally using `yarn start` in both `server` and `web`:
* Logged in
* Logged out
* Created request
* Confirmed request appeared in list

Prod build is currently untested - I'll get that deployed to S3 to verify.

sdunster:
Deployed now to https://web.callout.nsws.es

PR #253, opened 2018-04-03
Automatically deploy web to S3

sdunster:
* New travis build stage for deploying web
* `pip install awscli`
* `aws s3 sync dist s3://....`


sdunster:
Based off #253 - will rebase when it lands.
Tested full CI run through here:  https://travis-ci.org/NSWSESMembers/availability-poc/jobs/361506531
The only difference between that commit and this is that that one was configured to allow travis to build `sdunster.deploy_web`.

PR #254, opened 2018-04-03
Event Response Modals refactor

OSPFNeighbour:
Minor refactor to make modals written for this page more generic.

Added some smarts around making the ETA question say the name of the selected destination.


OSPFNeighbour:
@sdunster I looked to replace some of the function calls changed from () => funct to just funct but it would cause the whole app to hang when the page loaded. very strangle.

PR #262, opened 2018-04-19
Tags changes

OSPFNeighbour:
Changed creator for tags to remove user and group

Changed test data to include tags under group

Added more tags to test data

### DIFF
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -52,23 +52,14 @@ export const getCreators = (models) => {
       locationLongitude,
     }),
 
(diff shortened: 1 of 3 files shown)


server/src/creators.js
```diff
@@ -98,6 +89,7 @@ export const getCreators = (models) => {
     },
 
     group: ({ organisation, name, icon, tags, users }) => {
+      console.log('tag', tags);
```
sdunster: plz no

OSPFNeighbour: ffs i thought i got them all out

PR #263, opened 2018-04-19
Pull to Refresh Features

sjmcculloch:
Added pull to refresh feature for home & availability screens. 

This will allow a user to enter a new request in the web application and a user on the mobile can see it by pulling down on either of these screens.

Eventually will be automatic via push subscription. 

image

### DIFF
--- a/client/src/screens/availability/Index.js
+++ b/client/src/screens/availability/Index.js
@@ -35,6 +35,10 @@ class Index extends Component {
     this.props.navigation.navigate('Detail', { id: item.id });
   };
 
+  onRefresh = () => {
+    this.props.refetch();
+  }
+
   renderItem = ({ item }) => (
     <ListItem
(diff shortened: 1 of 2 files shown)


client/src/screens/availability/Index.js
```diff
@@ -80,6 +80,7 @@ class Index extends Component {
           keyExtractor={item => item.id}
           renderItem={this.renderItem}
           refreshing={this.props.networkStatus === 4}
+          onRefresh={() => this.props.refetch()}
```
sdunster: I think this is best done with a bind in init but I won't block you

sdunster: actually you might not even need to bind it now that I think about it:
```
onRefresh={this.props.refetch}
```
^ that might be enough because the prop provider should be responsible for binding

PR #264, opened 2018-04-19
tags not returning on group creation fixed

OSPFNeighbour:
...


OSPFNeighbour:
Was not fine. changed to the correct reload function as to return the original instance and not fetch a new one.

PR #265, opened 2018-04-19
word change on tag modal

OSPFNeighbour:
"Back" => "Save & Back"

screenshot 2018-04-19 17 31 52


sdunster:
I don't think you need to say "back" - it's obvious the dialog will close. As such I'd suggest just using "Save".

OSPFNeighbour:
Thx Sam, I agree.

PR #266, opened 2018-04-19
Make component out of set response

sdunster:
* New generic (stateful) `EditResponse` generic component that accepts a
  callback when completed so it can be plugged in and used wherever.
* New concrete wrapper `EditResponse` that implements the navigation part to
  push this onto the existing `StackEvents`.
* Later on we can add a `AddResponse` that uses the same generic component
  but instead displays a global modal (over the top of anything) and then
  routes directly to the event detail screen upon close.


sdunster:
Tested pretty extensively on iOS.

sdunster:
Pretty sure that lint error is bogus.. I silenced it.

PR #270, opened 2018-04-20
Re-organise the navigation system

sdunster:
* New hierachy of navigators under `navigation/`
* Remove react-navigation redux integration and go back to letting it manage
  state itself. It's too hard to work with and I don't think we need the
  complexity. The only downside is that users might not return to the same spot
  in the app after a re-launch but we always had that disabled anyway.
* Upgrade react-navigation
* Add loading interstitial that prevents loading of anything else until the app
  is initialised and we go to either auth or the main app depending on whether
  the user is authenticated or not.


sdunster:
This is still pretty raw so treat it as a preview. It works but it's a bit of a mess. I'll get it tidied up soon.

OSPFNeighbour:
While you are at it can you fix the home icon. its been pissing me off for months. For some reason we are only in a duplex sized house :p

screenshot 2018-04-20 19 57 19

client/src/navigation/AuthStackNavigator.js
```diff
@@ -1,15 +1,13 @@
 import { StackNavigator } from 'react-navigation';
 
-import Index from './Index';
-import SignIn from './SignIn';
-import SignUp from './SignUp';
-
-import NavOptions from '../../config/NavOptions';
+import { Home, SignIn, SignUp } from '../screens/auth';
```
OSPFNeighbour: I cant get this branch to build. This line causes


```
                                                                               Bundling `index.js`  [development, non-minified]  0.0% (1/777), failed.% (1/777)
error: bundling failed: UnableToResolveError: Unable to resolve module `../screens/auth` from `/Users/ttdykes/git/availability-poc/client/src/navigation/AuthStackNavigator.js`: could not resolve `/Users/ttdykes/git/availability-poc/client/src/screens/auth' as a file nor as a folder
```

client/src/navigation/AvailabilityStackNavigator.js
```diff
@@ -1,14 +1,12 @@
 import { StackNavigator } from 'react-navigation';
 
-import Index from './Index';
-import Detail from './Detail';
-import Requests from './Requests';
-import NavOptions from '../../config/NavOptions';
+import { Root, Detail, Requests } from '../screens/availability';
```
OSPFNeighbour: unresolvable path

client/src/navigation/EventsStackNavigator.js
```diff
@@ -1,18 +1,15 @@
 import { StackNavigator } from 'react-navigation';
 
-import Events from './Events';
-import Event from './Detail';
-import EditResponse from './EditResponse';
-import EventUsers from './EventUsers';
-import NavOptions from '../../config/NavOptions';
+import { Events, Detail, EditResponse, EventUsers } from '../screens/events';
```
OSPFNeighbour: unresolvable path

client/src/navigation/GroupsStackNavigator.js
```diff
@@ -1,15 +1,12 @@
 import { StackNavigator } from 'react-navigation';
 
-import Index from './Index';
-import GroupDetails from './GroupDetails';
-import NewGroup from './NewGroup';
-import SearchGroup from './SearchGroup';
-import NavOptions from '../../config/NavOptions';
+import { Root, GroupDetails, SearchGroup, NewGroup } from '../screens/groups';
```
OSPFNeighbour: unresolvable path

client/src/navigation/HomeStackNavigator.js
```diff
@@ -1,12 +1,12 @@
 import { StackNavigator } from 'react-navigation';
 
-import Index from './Index';
-import NavOptions from '../../config/NavOptions';
+import { Root } from '../screens/home';
```
OSPFNeighbour: unresolvable path

OSPFNeighbour:
Slide transition between stacks seems to render a blank screen during transition for the first time, subsequent slides render ok.

clicking button to move between stacks seems to be fine.

sdunster:
yeah its still pretty raw as I warned. I didn't get the unresolvable path problems though so that's weird.

OSPFNeighbour:
Would it be because the index file in the folder had a capital I?

sdunster:
oh yeah that'd do it. I removed those files and then re-created new ones so it works fine on my computer. Your case-insensitive filesystem would not have applied the changes correctly. Either rename the files locally or re-clone the entire repo. For future reference: never create `Index.js` files - terrible idea :P

sdunster:
are you sure that will fix it? I'll try it.

OSPFNeighbour:
The internet says it will. I think it was written for this sort of thing
where git can't tell something was renamed due to file system.

sdunster:
* fix some lint
* make sure there is an explicit rename from `Index.js` to `index.js` so case insensitive file system devs dont run into issues.

sdunster:
* make screen names match their filenames

sdunster:
* Rename navigators to match their filename.
* We now push screens onto the home navigator when a user taps on a event or
  schedule. As such, pressing back returns users to the home/notifications
  screen.
* Split home screen `<ListItem>` wrappers into their own components:
  * `<EventListItem>`
  * `<ScheduleListItem>`

PR #271, opened 2018-04-21
optimisticResponse for createTimeSegment

OSPFNeighbour:
Makes the UI a little more responsive by estimating what the server will return and warming up the cache

### DIFF
--- a/client/src/screens/schedules/Detail.js
+++ b/client/src/screens/schedules/Detail.js
@@ -170,6 +170,7 @@ class Detail extends Component {
 
   getTimeSegments = (day) => {
     const schedule = this.getSchedule();
+
     if (day !== undefined) {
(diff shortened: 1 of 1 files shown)


OSPFNeighbour:
The optimisticResponse data must look the same as the answer you are
optimistically expecting, so we need the id. given that we don't know it
yet it just needs to be unique in the cache so -1 is a safe bet.

Once the real data returns the cache will sort it's self out.

OSPFNeighbour:
Further to this, if we assume all negative IDs are optimistic we can have
the UI render them differently to give the user some feedback that it's
still loading

sdunster:
You've confirmed the data gets wiped upon refresh moments later?

OSPFNeighbour:
Yes, the cache is managed correctly.

PR #280, opened 2018-04-22
Set light-content status bar on iOS

sdunster:
* Finally this works!
* We can override the color by adding additional `<StatusBar>` elements on
  other screens.


OSPFNeighbour:
Screen shot pls.

sdunster:
screen shot 2018-04-22 at 02 08 56

PR #281, opened 2018-04-22
Disable console.log in prod app

OSPFNeighbour:
'The internet' says this is the easy way to do it.

Should give a minor performance increase.

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -17,7 +17,7 @@ import { GRAPHQL_ENDPOINT } from './config';
 
 function noop() {}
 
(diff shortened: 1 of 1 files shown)


sdunster:
I really doubt this is the cause of the performance problems you have been talking about. I believe it's useful to leave these enabled in prod, especially if bugsnag gathers them - it might help us try to diagnose tricky bugs in prod.

OSPFNeighbour:
Yea Sam I tend to agree, this was just the highest rated fix for performance issues when changing tabs in react

PR #282, opened 2018-04-22
Introduce 'urgent' events and modal for response

sdunster:
* We display a random event as urgent (just for testing)
* Urgent events will be events that meet certain criteria (ie rescue) that a
  user is able to respond to but has not yet recorded a response. Details tbc.
* Urgent events are shown on the home screen in a different colour and when
  tapped launch directly into the response flow.
* This repsonse flow is the same thing that will show in-app when you tap a
  push notification.
* Had to add a new `ModalNavigator` which sits between `RootNavigator`
  (`SwitchNavigator`) and `MainNavigator` (`TabNavigator`).
* The implementation is messy because I could not get `react-navigation` to
  navigate to a route defined in the `ModalNavigator` without passing it's
  navigation prop all the way down to the home root screen.

### DIFF
--- a/client/src/components/Events/EventListItem.js
+++ b/client/src/components/Events/EventListItem.js
@@ -5,14 +5,15 @@ import { ListItem } from '../List';
 
 // this component is intended to be used to display basic information about an event in a
 // <FlatList> and go to event detail view when tapped
-const EventListItem = ({ event, onPress }) => (
+const EventListItem = ({ event, urgent, onPress }) => (
   <ListItem
     title={event.name}
     bold
     subtitle={event.details}
     subtitleEllipsis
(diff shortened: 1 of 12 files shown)


sdunster:
screen shot 2018-04-22 at 04 10 39
screen shot 2018-04-22 at 04 10 41

sdunster:
also I tried to use `<SafeAreaView>` to not collide with the status bar but it didn't work. I'll tidy it up in a future commit.

OSPFNeighbour:
Think you need to add another set of test data for a non urgent event (like a high priority storm job) to make this concept more obvious.

I would add a priority tag to the event model and use that as the flag for this action.

PR #285, opened 2018-04-22
Suppress react warning relating to missing event response detail

sdunster:
* The view works fine without it set and I think it's normal for it to
  sometimes be unset so lets remove the `isRequired` PropType.


sdunster:
screen shot 2018-04-22 at 15 49 39

sjmcculloch:
lgtm

PR #290, opened 2018-04-23
Use built-in text truncation props instead of custom stuff

sdunster:
* Use `<Text>`'s `numberOfLines` prop.
* Implement into our custom `<ListItem>` with separate configuration for
  title vs subtitle.
* Make Availability and Events tab use the component ListItems we made for the
  home tab.
* We use 2 title + 3 subtitle maximum for urgent events. Regular events are
  1 + 2 as are schedules.


sdunster:
screen shot 2018-04-22 at 17 23 36

OSPFNeighbour:
Nice

PR #293, opened 2018-04-23
Increase tab bar font size slightly for iOS

sdunster:


sdunster:
before:
screen shot 2018-04-22 at 18 07 16

after:
screen shot 2018-04-22 at 18 08 20

PR #299, opened 2018-04-23
Prefix `Root` components names for better debugging

sdunster:
* Rename `Root` components so it is more obvious which is broken in stack
  traces.


client/src/screens/burger/Root.js
```diff
@@ -35,7 +35,7 @@ const updateUserProfileMutation = graphql(UPDATE_USERPROFILE_MUTATION, {
   }),
 });
 
-class Root extends Component {
+class BurgerRoot extends Component {
```
OSPFNeighbour: lol

PR #304, opened 2018-04-23
removed reload button from home loading view

OSPFNeighbour:
Yeap....


OSPFNeighbour:
do we need something?

PR #305, opened 2018-04-23
Availability updates for UI display and bug fixes for time selection

sjmcculloch:
- Added message for ongoing schedules in lieu of date selection
- Added scrollview to data entry for long entry areas
- Fixed icons in availability status legend
- Align timeSegment boxes to a fixed width of 48%

image

### DIFF
--- a/client/src/components/DateTime/styles.js
+++ b/client/src/components/DateTime/styles.js
@@ -76,9 +76,13 @@ const styles = StyleSheet.create({
     color: Colors.bdMain,
   },
   timeSelectContainer: {
+    width: '100%',
     flexDirection: 'row',
-    justifyContent: 'space-around',
     flexWrap: 'wrap',
+  },
+  timeSelectChild: {
+    width: '48%',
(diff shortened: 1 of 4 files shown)


client/src/components/DateTime/styles.js
```diff
@@ -76,10 +76,14 @@ const styles = StyleSheet.create({
...
   },
   timeSelectContainer: {
+    width: '100%',
     flexDirection: 'row',
-    justifyContent: 'space-around',
     flexWrap: 'wrap',
   },
+  timeSelectChild: {
+    width: '48%',
+    margin: '1%',
```
sdunster: i'm sure this will bite us at some point - I feel like this kinda stuff never works in CSS so it probably doesn't work here either. but it'll do for now.

sjmcculloch: I couldn't find an effective way other than this to lay out the timeSegment tiles - given the length of titles are different. We can revisit in the future.

PR #312, opened 2018-04-24
Increase client test coverage

sdunster:
* Coverage for all GraphQL queries/mutations/fragments
* Some extra components
* auth selector
* some other simple files


sdunster:
de-lint

PR #313, opened 2018-04-25
More client test coverage

sdunster:
* Some more low-hanging fruit in components
* Fixed a couple of minor bugs - will highlight in PR

### DIFF
--- a/client/src/components/Button/ButtonBox.js
+++ b/client/src/components/Button/ButtonBox.js
@@ -35,10 +35,10 @@ const ButtonBox = ({ text, subtext, onPress, selected, selectedColor, disabled }
   }
 
   return (
-    <TouchableOpacity style={buttonStyleBox} onPress={() => onPress()}>
+    <TouchableOpacity style={buttonStyleBox} onPress={onPress}>
       <Text style={textStyleBox}>{text}</Text>
       {subtext && <Text style={subtextStyleBox}>{subtext}</Text>}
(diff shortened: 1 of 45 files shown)


client/src/components/Button/ButtonBox.js
```diff
@@ -35,10 +35,10 @@ const ButtonBox = ({ text, subtext, onPress, selected, selectedColor, disabled }
   }
 
   return (
-    <TouchableOpacity style={buttonStyleBox} onPress={() => onPress()}>
+    <TouchableOpacity style={buttonStyleBox} onPress={onPress}>
       <Text style={textStyleBox}>{text}</Text>
       {subtext && <Text style={subtextStyleBox}>{subtext}</Text>}
-      {selected === false && <Icon size={18} name="times-circle" color="white" />}
+      {selected !== true && <Icon size={18} name="times-circle" color="white" />}
```
sdunster: I think this is misleading - you can't get this `<Icon>` to show unless you specify explicitly `<ButtonBox selected={false} />`. I would expect `<ButtonBox />` with no `selected` param specified to show this icon.

sdunster: The only usage of this currently is this:
```
selected={segment.status !== ''}
```
which would work before and after this change.
@sjmcculloch I'm not going to wait for your review but you should confirm this is cool at some point.

sdunster:
de-lint

PR #318, opened 2018-04-25
Remove unused components

sdunster:
* Remove `<ListItemAction>`
* Remove `<TextInputRow>`


sdunster:
@OSPFNeighbour @sjmcculloch can you both confirm you don't intend to use these?
I tend to think both of these components can be re-made using other existing components rather than special-snowflakes/copy-pasta.

OSPFNeighbour:
They are left over from the first version of Scotts schedule input. I would remake something if i wanted something given my increased react skills.

PR #320, opened 2018-04-28
Group Details Page Rewrite

OSPFNeighbour:
Bringing the group details page into the same theme as the rest of the app.

- Total rewrite of the page to use existing UI modules
- Added new module for User ListItem inline with existing ListItems

screenshot 2018-04-28 19 32 48

### DIFF
--- a/client/src/components/Users/index.js
+++ b/client/src/components/Users/index.js
@@ -1,0 +1,4 @@
+import UserListItem from './UserListItem';
+import styles from './styles';
+
+export { UserListItem, styles };
--- a/client/src/screens/groups/GroupDetails.js
+++ b/client/src/screens/groups/GroupDetails.js
@@ -1,253 +1,35 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import {
-  FlatList,
   Text,
-  ActivityIndicator,
-  TouchableHighlight,
-  View,
-  Button,
   Alert,
 } from 'react-native';
 import { graphql, compose } from 'react-apollo';
-import Icon from 'react-native-vector-icons/FontAwesome';
 import { connect } from 'react-redux';
 import _ from 'lodash';
-import moment from 'moment';
 
-import distantFuture from '../../constants';
-import { Container } from '../../components/Container';
-import { extendAppStyleSheet } from '../style-sheet';
+import { Center, Container, Holder } from '../../components/Container';
+import { Progress } from '../../components/Progress';
+import { Segment } from '../../components/Segment';
+
+
 import SEARCH_GROUP_QUERY from '../../graphql/search-group.query';
 import JOIN_GROUP_MUTATION from '../../graphql/join-group.mutation';
 import LEAVE_GROUP_MUTATION from '../../graphql/leave-group.mutation';
 import CURRENT_USER_QUERY from '../../graphql/current-user.query';
 
-const styles = extendAppStyleSheet({
-  groupContainer: {
-    flex: 1,
-    flexDirection: 'row',
-    alignItems: 'center',
-    backgroundColor: 'white',
-    borderBottomColor: '#eee',
-    borderBottomWidth: 1,
-    paddingHorizontal: 12,
-    paddingVertical: 8,
-  },
-  headerContainer: {
-    flex: 1,
-    flexDirection: 'row',
-    borderBottomColor: '#eee',
-    borderBottomWidth: 1,
-    paddingHorizontal: 12,
-    paddingVertical: 8,
-  },
-  headerTextContainer: {
-    flex: 1,
-    flexDirection: 'column',
-    paddingLeft: 6,
-  },
-  sectionTextContainer: {
-    flex: 1,
-    flexDirection: 'column',
-    paddingLeft: 6,
-    backgroundColor: '#eee',
-  },
-  headerName: {
-    fontWeight: 'bold',
-    fontSize: 18,
-  },
-  groupName: {
-    fontWeight: 'bold',
-    flex: 0.7,
-  },
-  groupTitleContainer: {
-    flexDirection: 'row',
-  },
-  groupTextContainer: {
-    flex: 1,
-    flexDirection: 'column',
-    paddingLeft: 6,
-  },
-  eventContainer: {
-    flex: 1,
-    flexDirection: 'row',
-    alignItems: 'center',
-    backgroundColor: 'white',
-    borderBottomColor: '#eee',
-    borderBottomWidth: 1,
-    paddingHorizontal: 12,
-    paddingVertical: 8,
-  },
-  eventName: {
-    fontWeight: 'bold',
-    flex: 0.7,
-  },
-  eventTitleContainer: {
-    flexDirection: 'row',
-  },
-  eventTextContainer: {
-    flex: 1,
-    flexDirection: 'column',
-    paddingLeft: 6,
-  },
-  eventText: {
-    color: '#8c8c8c',
-  },
-  scheduleContainer: {
-    flex: 1,
-    flexDirection: 'row',
-    alignItems: 'center',
-    backgroundColor: 'white',
-    borderBottomColor: '#eee',
-    borderBottomWidth: 1,
-    paddingHorizontal: 12,
-    paddingVertical: 8,
-  },
-  scheduleName: {
-    fontWeight: 'bold',
-    flex: 0.7,
-  },
-  scheduleTitleContainer: {
-    flexDirection: 'row',
-  },
-  scheduleTextContainer: {
-    flex: 1,
-    flexDirection: 'column',
-    paddingLeft: 6,
-  },
-  scheduleText: {
-    color: '#8c8c8c',
-  },
-});
+import GroupDetailsHeader from './components/GroupDetailsHeader';
+import GroupDetailsList from './components/GroupDetailsList';
 
-const GroupHeader = (props) => {
-  const { name, tags } = props.group;
-  const tagWord = tags.map(elem => `#${elem.name}`).join(' ');
-
-  return (
-    <View>
-      <View style={styles.headerContainer}>
-        <View style={styles.headerTextContainer}>
-          <Text style={styles.headerName} numberOfLines={1}>{name}</Text>
-          <Text style={styles.headerDetail} numberOfLines={3}>{tagWord}</Text>
-        </View>
-      </View>
-    </View>
-  );
-};
-GroupHeader.propTypes = {
-  group: PropTypes.shape({
-    name: PropTypes.string,
-    tags: PropTypes.arrayOf(
-      PropTypes.shape({
-        id: PropTypes.number.isRequired,
-        name: PropTypes.string.isRequired,
-      }),
-    ),
-  }),
-};
-
-const SectionHeader = (props) => {
-  const { section } = props;
-
-  return (
-    <View>
-      <View style={styles.headerContainer}>
-        <View style={styles.sectionTextContainer}>
-          <Text style={styles.headerName} numberOfLines={1}>{section}</Text>
-        </View>
-      </View>
-    </View>
-  );
-};
-SectionHeader.propTypes = {
-  section: PropTypes.string,
-};
-
-const MembersDisplay = (props) => {
-  const { id, displayName } = props.member;
-  return (
-    <TouchableHighlight key={id}>
-      <View style={styles.groupContainer}>
-        <Icon name="user" size={24} color="blue" />
-        <View style={styles.groupTextContainer}>
-          <View style={styles.groupTitleContainer}>
-            <Text style={styles.groupName} numberOfLines={1}>{displayName}</Text>
-          </View>
-        </View>
-      </View>
-    </TouchableHighlight>
-  );
-};
-MembersDisplay.propTypes = {
-  member: PropTypes.shape({
-    id: PropTypes.number.isRequired,
-    displayName: PropTypes.string.isRequired,
-  }),
-};
-
-const EventsDisplay = (props) => {
-  const { id, name, details } = props.event;
-  return (
-    <TouchableHighlight key={id}>
-      <View style={styles.eventContainer}>
-        <Icon name="bullhorn" size={24} color="blue" />
-        <View style={styles.eventTextContainer}>
-          <View style={styles.eventTitleContainer}>
-            <Text style={styles.eventName} numberOfLines={1}>{name}</Text>
-          </View>
-          <Text style={styles.eventText} numberOfLines={1}>{details}</Text>
-        </View>
-      </View>
-    </TouchableHighlight>
-  );
-};
-EventsDisplay.propTypes = {
-  event: PropTypes.shape({
-    id: PropTypes.number.isRequired,
-    name: PropTypes.string.isRequired,
-  }),
-};
-
-const ScheduleDisplay = (props) => {
-  const { id, name, details, startTime, endTime } = props.schedule;
-  let timeText = '';
-  if (startTime === 0 && endTime === distantFuture) {
-    timeText = 'Perpetual Schedule';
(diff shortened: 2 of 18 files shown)


OSPFNeighbour:
Im not in love with the layout of the top bar and the sector divider but its all I could come up with

client/src/screens/groups/GroupDetails.js
```diff
@@ -3,252 +3,89 @@ import React, { Component } from 'react';
...
-};
-ScheduleDisplay.propTypes = {
-  schedule: PropTypes.shape({
-    id: PropTypes.number.isRequired,
-    name: PropTypes.string.isRequired,
-    details: PropTypes.string,
-    startTime: PropTypes.number.isRequired,
-    endTime: PropTypes.number.isRequired,
-  }),
+const GroupDetailsList = ({ items, render, networkStatus, refetch }) => (
```
sdunster: some tips to improve this:
* change this to a function and make it return the React components
  ```
  const GroupDetailsList = ({ items, render, networkStatus, refetch }) => {
    return <Container>...</Container>
  }
  ```
* define `renderItem` as a named function so that we can pull the complicated logic out:
  ```
  const renderItem = ({item}) => {
    ...
  }
  ```
* change from nested ternary (like the linter suggests this is probably a bad idea™) to if/else-if:
  ```
  if (mode === 'user') {
    return <UserListItem ... />
  ...
  ```
* I believe you should avoid using the name of components to select them, unless you're actually going to use the component itself. If you've gotta do translation anyway it's less confusing to the reader. If I see 'UserListItem' as a string I'm going to assume you're dynamically looking up the component name, which you're not.
* I think you should avoid using the name `render` because this name is usually used for the function that renders a component. I suggest `mode`.
* Put this component in it's own file separate from the screen.

client/src/screens/groups/GroupDetails.js
```diff
@@ -3,252 +3,89 @@ import React, { Component } from 'react';
...
   static navigationOptions = () => ({
     title: 'Group Details',
   });
 
+  state = {
+    selectedIndex: 0,
+  };
+
+  componentDidMount() {
+    this.props.navigation.setParams({ navIconJoinLeave: 'sign-out' });
```
sdunster: what does this do? add a comment explaining it

client/src/screens/groups/GroupDetails.js
```diff
@@ -325,36 +138,72 @@ class GroupDetails extends Component {
...
-      _.map(group.users, r => ['user', r]),
-      _.map(['Events'], r => ['section', r]),
-      _.map(group.events, r => ['event', r]),
-      _.map(['Schedules'], r => ['section', r]),
-      _.map(group.schedules, r => ['schedule', r]),
-    );
-    // render group stuff
-
     const memberAlready = _.some(group.users, g => g.id === me.id);
+    const tags = group.tags.map(elem => `#${elem.name}`).join(', ');
```
sdunster: we could put this in a utils file or something so we can reuse it across the app

client/src/screens/groups/GroupDetails.js
```diff
@@ -325,36 +138,72 @@ class GroupDetails extends Component {
...
+          </View>
+          <View style={styles.textContainer}>
+            <Text style={styles.titleTextBold} numberOfLines={1}>
+              {group.name}
+            </Text>
+            <Text style={styles.subtitleText} numberOfLines={1}>
+              {tags !== '' ? tags : 'No Tags'}
+            </Text>
+          </View>
+          <TouchableOpacity
```
sdunster: this should be a component

client/src/screens/groups/GroupDetails.js
```diff
@@ -325,36 +138,72 @@ class GroupDetails extends Component {
...
+            <Icon style={styles.icon} name={memberAlready ? 'sign-out' : 'sign-in'} size={25} />
+            <Text style={styles.leaveGroupText}>{memberAlready ? 'Leave Group' : 'Join Group'}</Text>
+          </TouchableOpacity>
+        </View>
+        <Holder margin transparent>
+          <Segment
+            values={['Users', 'Schedules', 'Events']}
+            handleIndexChange={this.handleIndexChange}
+            selectedIndex={this.state.selectedIndex}
+            refetch={this.onRefresh}
```
sdunster: what does this do?

client/src/screens/groups/GroupDetails.js
```diff
@@ -325,36 +138,72 @@ class GroupDetails extends Component {
...
+          />
+        </Holder>
+        {this.state.selectedIndex === 0 && (
+        <GroupDetailsList
```
sdunster: how about we make this component take the entire `group` object and then pick users, schedules or events based upon the `render` parameter (although I recommend `mode` for reasons specified above)? This way you only need to define GroupDetailsList once. I don't think it matters that you pass down props you're not going to use for each particular render - the cost of passing some references is minimal.
```
<GroupDetailsList
  group={group}
  mode={['users','schedules','events'][this.state.selectedIndex]}
  networkStatus={networkStatus}
  refetch={this.onRefresh}
/>
```

client/src/screens/groups/GroupDetails.js
OSPFNeighbour: what do you gain by making something into a function and returning the component, compared to keeping it how ive got it... { with a return vs (... (I dont know what this is called.)

client/src/screens/groups/GroupDetails.js
OSPFNeighbour: worked around this in a smarter way

client/src/screens/groups/GroupDetails.js
OSPFNeighbour: this should not be in there. nice catch

client/src/screens/groups/GroupDetails.js
sdunster: it will enable you to do the other suggestions. Both are functions - the latter lets you have one or more statements whereas the former only allows a single expression.

client/src/components/Users/index.js
```diff
@@ -0,0 +1,4 @@
+import UserListItem from './UserListItem';
+
+// eslint-disable-next-line import/prefer-default-export
```
sjmcculloch: Can you add a styles.js to the same folder to avoid this and also setup same pattern as other components.

client/src/screens/groups/GroupDetails.js
sjmcculloch: Segment is the component with the 3 selectable areas - this updates a piece of state to conditionally render content in the parent components.

client/src/screens/groups/styles.js
```diff
@@ -6,6 +6,65 @@ const FONT_SIZE = 16;
 const HIGHLIGHT_COLOR = 'rgba(0,118,255,0.9)';
```
sjmcculloch: Can we eventually move this to /themes/colors.js

client/src/screens/groups/styles.js
```diff
@@ -6,6 +6,65 @@ const FONT_SIZE = 16;
...
+    marginLeft: 10,
+    marginRight: 10,
+    marginTop: 6,
+    marginBottom: 0,
+  },
+  leaveGroupText: {
+    fontSize: 7,
+    includeFontPadding: false,
+    flex: 0,
+    color: '#6B6B6B',
```
sjmcculloch: In future - can we move these colors to themes/colors.js

client/src/screens/groups/components/GroupDetailsRenderItem.js
```diff
@@ -0,0 +1,46 @@
...
+  Alert,
+} from 'react-native';
+
+import EventListItem from '../../../components/Events/EventListItem';
+import ScheduleListItem from '../../../components/Schedules/ScheduleListItem';
+import UserListItem from '../../../components/Users/UserListItem';
+
+const GroupDetailsRenderItem = ({ item }) => {
+  // eslint-disable-next-line no-underscore-dangle
+  switch (item.__typename) { // GQL typename from Schema
```
sdunster: neat

client/src/screens/groups/components/GroupDetailsRenderItem.js
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+import {
+  Alert,
+} from 'react-native';
+
+import EventListItem from '../../../components/Events/EventListItem';
+import ScheduleListItem from '../../../components/Schedules/ScheduleListItem';
+import UserListItem from '../../../components/Users/UserListItem';
+
+const GroupDetailsRenderItem = ({ item }) => {
```
sdunster: why don't you just put this function at the top of `GroupDetailsList`? Either in the render function or on the line before it? This isn't a component so it feels weird to have it as a separate file, especially when only used with `GroupDetailsList`.

client/src/screens/groups/components/GroupDetailsRenderItem.js
```diff
@@ -0,0 +1,46 @@
...
+  // eslint-disable-next-line no-underscore-dangle
+  switch (item.__typename) { // GQL typename from Schema
+    case 'User':
+      return (
+        <UserListItem
+          user={item}
+      // TODO: Deep link
+          onPress={() => Alert.alert('PLACEHOLDER', 'Deeplink to user details')}
```
sdunster: when you do get around to implementing this I think it'd be best to do it as a callback that provides the type and the id so the caller can decide how to route it.

client/src/screens/groups/components/GroupDetailsRenderItem.js
```diff
@@ -0,0 +1,46 @@
...
+    // TODO: Deep link
+          onPress={() => Alert.alert('PLACEHOLDER', 'Deeplink to event details')}
+        />
+      );
+  }
+};
+
+GroupDetailsRenderItem.propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  item: PropTypes.object.isRequired,
```
sdunster:
```
__typename: PropTypes.string.isRequired,
```

client/src/screens/groups/components/styles.js
```diff
@@ -0,0 +1,281 @@
+import { StyleSheet } from 'react-native';
+
+const PADDING = 8;
+const BORDER_RADIUS = 5;
+const FONT_SIZE = 16;
+const HIGHLIGHT_COLOR = 'rgba(0,118,255,0.9)';
+
+const styles = StyleSheet.create({
```
sdunster: there's a lot of stuff in here.. is all of this actually used?

client/src/screens/groups/components/GroupDetailsRenderItem.js
OSPFNeighbour: Agreed. everything is there for the link to work just need the plumbing

client/src/screens/groups/GroupDetails.js
OSPFNeighbour: looks like its not a valid prop. will remove.

client/src/components/Users/index.js
OSPFNeighbour: it currently has no style, do we need a style sheet even when it has no style?

client/src/screens/groups/components/GroupDetailsHeader.js
```diff
@@ -0,0 +1,49 @@
...
+    </View>
+    <View style={styles.textContainer}>
+      <Text style={styles.titleTextBold} numberOfLines={1}>
+        {groupName}
+      </Text>
+      <Text style={styles.subtitleText} numberOfLines={1}>
+        {tags !== '' ? tags : 'No Tags'}
+      </Text>
+    </View>
+    <TouchableOpacity
```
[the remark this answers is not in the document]
OSPFNeighbour: Sorry, I totally misunderstood what you where trying to do, Ill rewrite this to be more compact.

client/src/screens/groups/components/GroupDetailsRenderItem.js
OSPFNeighbour: Done

client/src/screens/groups/components/GroupDetailsRenderItem.js
OSPFNeighbour: its within the object

client/src/screens/groups/components/styles.js
OSPFNeighbour: no. a lot of it isnt

client/src/screens/groups/components/GroupDetailsHeader.js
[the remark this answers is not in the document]
OSPFNeighbour: @sdunster as in split this up even further?

client/src/components/Users/index.js
sdunster: it's probably not a terrible idea even if it is an empty dict

client/src/screens/groups/components/GroupDetailsHeader.js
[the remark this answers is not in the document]
sdunster: well yeah - make a component that is "IconButton" even if it is specific to users. Even better make it generic enough that we can re-use it across the app. Remember that special snowflakes aren't just bad for the code (extra work to maintain, need to have tests etc) but it's also bad for UI design (users don't recognise it as a button or don't know what it does because it's only used in this one spot in the app). If we don't have something else that would work better here lets make it a component that we can use in others places.

client/src/screens/groups/components/GroupDetailsList.js
```diff
@@ -0,0 +1,77 @@
...
+} from 'react-native';
+
+import { Center, Container } from '../../../components/Container';
+
+import EventListItem from '../../../components/Events/EventListItem';
+import ScheduleListItem from '../../../components/Schedules/ScheduleListItem';
+import UserListItem from '../../../components/Users/UserListItem';
+
+
+const GroupDetailsRenderItem = ({ item }) => {
```
sdunster: nit: regular functions (that aren't supposed to 'appear' to be components or classes) should probably start with a lower case character (`groupDetailsRenderItem`)

client/src/screens/groups/components/GroupDetailsList.js
```diff
@@ -0,0 +1,77 @@
...
+    <FlatList
+      data={items[selectedIndex]}
+      ListHeaderComponent={() =>
+        (!items.length ? (
+          <Center>
+            <Text>Nothing here</Text>
+          </Center>
+        ) : null)}
+      keyExtractor={item => item.id}
+      renderItem={item => GroupDetailsRenderItem(item)}
```
sdunster: a function that calls another function with the same parameter and returns the value from the function it calls is basically a no-op:
```
renderItem={groupDetailsRenderItem}
```
^ this should work exactly the same and saves the creation of extra anonymous functions

client/src/screens/groups/components/GroupDetailsList.js
OSPFNeighbour: I always wondered what the rule was on that. TIL

client/src/screens/groups/components/GroupDetailsList.js
OSPFNeighbour: ty

OSPFNeighbour: im going to not action the button idea in this PR as there are a few places I can swap into a shared component that I will do a single PR for all of them

client/src/screens/groups/GroupDetails.js
OSPFNeighbour: agree. theres a few places that use it so ill open a new PR

client/src/screens/groups/styles.js
OSPFNeighbour: deleted them.. so that fixes that 👍

sdunster:
nice work

PR #321, opened 2018-04-28
Tests for DateTime component

OSPFNeighbour:
tenor

### DIFF
--- a/client/src/components/DateTime/__tests__/DateSelect.js
+++ b/client/src/components/DateTime/__tests__/DateSelect.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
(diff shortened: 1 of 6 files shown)


client/src/components/DateTime/__tests__/DateSelect.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import DateSelect from '../DateSelect';
+
+const noop = () => {};
+
+test('renders correctly', () => {
```
sdunster: if these are all just doing render/snapshot tests you should probably just put them all in the same `test()` like I did for the others I created - this seemed to be the recommended way from the jest docs.

PR #322, opened 2018-04-28
Tests for Modal component

OSPFNeighbour:
tenor 1

### DIFF
--- a/client/src/components/Modal/__tests__/Modal.js
+++ b/client/src/components/Modal/__tests__/Modal.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { Text } from 'react-native';
+import renderer from 'react-test-renderer';
+
+import Modal from '../Modal';
+
+const noop = () => {};
+
+const child = (
+  <Text>child</Text>
+);
+
+test('renders correctly', () => {
+  const tree1 = renderer.create(
+    <Modal
+      title="test"
+      closeModal={noop}
+      visible
+    >
+      {child}
+    </Modal>,
+  ).toJSON();
+  expect(tree1).toMatchSnapshot();
+});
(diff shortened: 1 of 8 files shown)


client/src/components/Modal/__tests__/Modal.js
```diff
@@ -0,0 +1,23 @@
...
+
+const noop = () => {};
+
+const child = (
+  <Text>test</Text>
+);
+
+test('renders correctly', () => {
+  const tree1 = renderer.create(
+    <Modal
```
sdunster: lol wut did you forget how to nest components?
```
<Modal title="test" closeModal={noop} visible>
  {child}
</Modal>
```

client/src/components/Modal/__tests__/NumberInputModal.js
```diff
@@ -0,0 +1,21 @@
...
+import renderer from 'react-test-renderer';
+
+import NumberInputModal from '../NumberInputModal';
+
+const noop = () => {};
+
+test('renders correctly', () => {
+  const tree1 = renderer.create(
+    <NumberInputModal
+      title="test"
```
sdunster: if you put different text in the attributes then we can more easily detect someone mixing prop names up. ie if you set both `title="test"` and `placeHolder="test"` and someone puts `{placeHolder}` where there is supposed to be `{title}` then the snapshot test wouldn't catch that bug. If the attributes are all different there would be a change that the snapshot test would catch. As such I recommend doing what I did for the others - for string props put the name of the prop as the test value.

client/src/components/Modal/__tests__/TextInputModal.js
```diff
@@ -0,0 +1,21 @@
...
+import renderer from 'react-test-renderer';
+
+import TextInputModal from '../TextInputModal';
+
+const noop = () => {};
+
+test('renders correctly', () => {
+  const tree1 = renderer.create(
+    <TextInputModal
+      title="test"
```
sdunster: `title="title"`

client/src/components/Modal/__tests__/Modal.js
OSPFNeighbour: ooh is that what it meant. duh.

PR #324, opened 2018-04-29
More Test Data

OSPFNeighbour:
Added a metro zone group & OCES group with some schedules to give demo data of how groups can be used

### DIFF
--- a/server/src/test-data/fixtures.js
+++ b/server/src/test-data/fixtures.js
@@ -227,7 +227,7 @@ export const GROUPS = [
 
 export const SCHEDULES = [
   {
-    name: 'Wagga Wagga OOA',
+    name: 'Wagga Wagga OOA availability',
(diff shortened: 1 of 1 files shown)


server/src/test-data/fixtures.js
```diff
@@ -203,11 +203,31 @@ export const GROUPS = [
...
+    tags: [
+      'NSW',
+    ],
+  },
 ];
 
 export const SCHEDULES = [
   {
-    name: 'Wagga Wagga OOA',
+    name: 'Wagga Wagga OOA availability',
```
sdunster: red flag any time I see a constant string appended/prepended to all entries.. I wonder if we should build it into the app so everything always says availability after it? Or do you see situations where we put something other than availability there? No changes required to this diff but we should discuss.

OSPFNeighbour: yes, totally agree. the UI should be handing this but I need a quick win for the Wednesday demo :P

PR #325, opened 2018-04-29
Refactor capabilities into tags

OSPFNeighbour:
Migration of the user capabilities data into Tag data.

Not trying to be better than the current capabilities or tags, just moving the code over. So I haven't added user tag handlers or screens yet.

The overall Apollo cache will be a bit happier now as we were missing ID on some data type so the cache would have been confused.

**Server**

Update to DB model
- remove capabilities
- add type to tags as ENUM (prob not a good idea long term)

Update of GQL schema
- type added
- filter to allow queries to filter on type

Update of test data & generator

Update to integration tests
- fixed from previous model to pre this PR model (was still expecting user and group to be passed)
- updated to support tag type

logic.js updated
- removed capabilities  
- query filter on tag name or type

**Client**

Update to Apollo GQL queries
- added id to anything that was missing it
- add tag type and tag filters
- reorder id to be first in all queries

Update to screens
- type added to props
- default types on queries added

### DIFF
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -5,7 +5,7 @@ import { DISTANT_FUTURE } from './constants';
 // returns a set of creators bound to the given models
 export const getCreators = (models) => {
   const {
-    Organisation, Group, User, Capability, Tag, Device, Event,
+    Organisation, Group, User, Tag, Device, Event,
     Schedule, TimeSegment, EventResponse, EventLocation, Message,
   } = models;
 
@@ -52,30 +52,18 @@ export const getCreators = (models) => {
       locationLongitude,
     }),
 
-    tag: ({ name, organisation }) => {
+    tag: ({ name, type, organisation }) => {
       if (!organisation || !organisation.id) {
         return Promise.reject(Error('Must pass organisation'));
+      }
+      if (!type) {
+        return Promise.reject(Error('Must pass type'));
       }
       return Tag.create({
         name,
+        type,
         organisationId: organisation.id,
       });
-    },
-
-    capability: ({ name, organisation, user }) => {
-      if (!organisation || !organisation.id) {
-        return Promise.reject(Error('Must pass organisation'));
-      }
-      if (!user || !user.id) {
-        return Promise.reject(Error('Must pass user'));
-      }
-      return Capability.create({
-        name,
-        organisationId: organisation.id,
-      }).then(capability =>
-        capability.addUser(user)
-          .then(() => user),
-      );
     },
 
     device: ({ uuid, user }) => {
@@ -107,7 +95,7 @@ export const getCreators = (models) => {
       ]).then(() => group.reload()));
     },
 
-    user: ({ id, username, password, email, displayName, version, organisation }) => {
+    user: ({ id, username, password, email, displayName, version, organisation, tags }) => {
       // it's fine for id to be left null/undefined
       if (!organisation || !organisation.id) {
         return Promise.reject(Error('Must pass organisation'));
@@ -121,7 +109,12 @@ export const getCreators = (models) => {
           email,
           version,
           organisationId: organisation.id,
-        }),
+        }).then(user => Promise.all([
+          tags && tags.map(
+            t => Tag.findById(t.id).then(foundTag => foundTag.addUser(user)),
+          ),
+        ]).then(() => user.reload()),
+        ),
       );
     },
 
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -108,9 +108,6 @@ export const getHandlers = ({ models, creators: Creators }) => {
       tags(user) {
         return user.getTags();
       },
-      capabilities(user) {
-        return user.getCapabilities();
-      },
       signup(args, ctx) {
         const { deviceUuid, email, username, password } = args.user;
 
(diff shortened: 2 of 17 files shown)


server/src/models.js
```diff
@@ -20,12 +20,12 @@ export const defineModels = (db) => {
...
 
-  const CapabilityModel = db.define('capability', {
-    name: { type: Sequelize.STRING },
-  });
-
   const TagModel = db.define('tag', {
     name: { type: Sequelize.STRING },
+    type: {
+      type: Sequelize.ENUM,
+      values: ['group', 'capability'],
```
sjmcculloch: Are we using "group" to mean "unit/region" in this context? It might be a bit confusing if we also call the type of the tag "group"

server/src/test-data/fixtures.js
```diff
@@ -22,25 +22,23 @@ export const DEFAULT_GROUP = 'NSW SES';
...
 export const ORG_NAME = 'NSW SES';
 
-export const CAPABILITIES = [
-  'Road Crash Rescue',
-  'Vertical Rescue',
-  'Flood Rescue',
-  'Chainsaw',
-];
 
 export const TAGS = [
```
sjmcculloch: These are the capabilities that I received from Mel:


Qualification in SAP | Roll Up Name in SAP
-- | --
Storm Ground | Storm and Water Damage - Ground
Storm Heights | Storm and Water Damage - Heights
Flood Rescue Operator Level 1 | Flood Rescue Operator Level 1
Flood Rescue Operator Level 2 | Flood Rescue Operator Level 2
Flood Rescue Boat Operator | Flood Rescue Boat Operator
Flood Rescue Operator Level 3 | Flood Rescue Operator Level 3
GLR | General Land Rescue Operator
GLR2 | TBC
CFR | Community First Responder Registration
VR | Vertical Rescue Operator
Chainsaw (L1 Cross Cut) | Chainsaw L1 (Cross Cut)
Chainsaw (L2 Intermediate Felling) | Chainsaw L2 (Intermediate Felling)
Land Search 1 | Land Search

The following qualifications are yet to be determined:
 
·        Chainsaw (L3)
·        Aviation
·        Land Search 2
·        Land Search 3 (Alpine)
·        First Tier Training (Foundations) – Future State: Job Ready
·        Foundations – Future State: Job Ready


IMIC001 | SES IM Incident Controller
IMPO001 | SES IM Planning Officer
IMPIO001 | SES IM Public Information Officer
IMOO001 | SES IM Operations Officer
IMLO001 | SES IM Logistics Officer

---

The ones you have are OK for now but if we could put some IMT roles in it would be helpful.

sjmcculloch:
And fix Travis 👍

server/src/models.js
sdunster: can you check whether this is actually making an enum type in the DB? If it is I caution against it. Having to do a schema change every time you want to add an extra option to an enum is a pita. A set of constant strings or integer ID that is looked up in a `js` file might be preferable. Remember to optimise for less effort rather than increased server efficiency.

server/src/creators.js
```diff
@@ -121,7 +109,12 @@ export const getCreators = (models) => {
           email,
           version,
           organisationId: organisation.id,
-        }),
+        }).then(user => Promise.all([
+          tags && tags.map(
+            t => Tag.findById(t.id).then(foundTag => foundTag.addUser(user)),
+          ),
+        ]).then(() => user.reload()),
```
sdunster: nice! you got promises all figured out!

server/src/logic.js
```diff
@@ -419,14 +416,17 @@ export const getHandlers = ({ models, creators: Creators }) => {
         return org.getUsers();
       },
       tags(org, args) {
-        if (args.filter) {
-          return org.getTags({ where: { name: { [Op.like]: `%${args.filter}%` } }, order: [['id', 'ASC']] });
+        if (args.nameFilter && !args.typeFilter) {
```
sdunster: I think we can do better than defining the queries twice... I'll write some pseudo-code, see if you can make it work:
```
const where = {};
if (args.nameFilter) {
  where.name = { [Op.like]: `%${args.nameFilter}%` };
}
if (args.typeFilter) {
  where.type = { [Op.like]: `%${args.typeFilter}%` };
}
return org.getTags({ where, order: [['id', 'ASC']] };
```
^ this has the benefit of defining the parts of the where clause only once, assuming it will gladly take an empty where (pretty sure it will). It also means that an empty filter results in a sorted list of tags (your current implementation does not).

server/src/test-data/fixtures.js
```diff
@@ -22,25 +22,37 @@ export const DEFAULT_GROUP = 'NSW SES';
...
-  'ISR',
-  'SSR',
-  'KMA',
-  'PAR',
-  'HLS',
-  'WOL',
-  'AUB',
-  'HAW',
-  'HOL',
+  { name: 'NSW', type: 'orgStructure' },
```
sdunster: instead of copy paste why don't you make the code work for you?
```
const makeTags = (rawTags) => {
  const results = [];
  rawTags.each((tags, type) => {
    tags.each((tag) => {
      results.push({ name: tag, type });
    });
  });
  return results;
});

export const TAGS = makeTags({
  orgStructure: [
    'SWR',
    'ISR',
    ...
  ],
  capability: [
    'Storm and Water Damage - Ground',
    'Storm and Water Damage - Heights',
  ],
});
```

OSPFNeighbour: Done. for the record when did it become so hard to walk over an object! you cant use .each on an object so the above didnt work. used...

```
const makeTags = (rawTags) => {
  const results = [];
  Object.entries(rawTags).forEach((entry) => {
    const [type, value] = entry;
    value.forEach((tag) => {
      results.push({ name: tag, type });
    });
  });
  return results;
};
```

server/src/logic.js
sdunster: it started as pseudo code and became real code, forgot to update comment.

PR #330, opened 2018-05-01
Travis Jest tweaks

OSPFNeighbour:
Should fix the stupidly slow jest snapshot test times we are currently getting on Travis but limiting the number of workers that Travis spawns (currently unlimited) by locking it down to 4

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -4,8 +4,8 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
-    "test": "TZ=UTC jest",
-    "test-and-coverage": "TZ=UTC jest && codecov -F unittests",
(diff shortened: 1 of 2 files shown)


OSPFNeighbour:
As per 

https://github.com/facebook/jest/issues/3855

and 

https://facebook.github.io/jest/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server

OSPFNeighbour:
Before:
Ran for 15 min

After:
Ran for 3 min 19 sec

sdunster:
fuck yeah nice fix! Thanks for attaching links to the advice :)

PR #333, opened 2018-05-02
Dashboard update

justrossthings:
updating dashboard to design: https://balsamiq.cloud/sitdcdu/peel1l3/r2321

TODO: Make sure schedule type is returned
TODO: Update link and add hide/show functionality to no capabilities callout

### DIFF
--- a/web/src/components/DisplayRequests.js
+++ b/web/src/components/DisplayRequests.js
@@ -45,10 +45,10 @@ const DisplayRequests = ({ classes, user }) => {
               {/* TODO: Make sure schedule object has this in the future */}
               <TableCell>{schedule.type}</TableCell>
               <TableCell>
-                {schedule.startTime === numbers.zero ? '-' : moment.unix(schedule.startTime).format('LLL')}
+                {schedule.startTime === numbers.distantPast ? '-' : moment.unix(schedule.startTime).format('LLL')}
               </TableCell>
               <TableCell>
-                {schedule.endTime === numbers.undefined
+                {schedule.endTime === numbers.distantFuture
                   ? '-'
                   : moment.unix(schedule.endTime).format('LLL')}
               </TableCell>
--- a/web/src/components/DisplayRequests.js
+++ b/web/src/components/DisplayRequests.js
@@ -30,7 +30,6 @@ const DisplayRequests = ({ classes, user }) => {
           <TableRow>
             <TableCell>Name</TableCell>
             <TableCell>Group</TableCell>
-            <TableCell>Type</TableCell>
             <TableCell>Start Date</TableCell>
             <TableCell>End Date</TableCell>
           </TableRow>
@@ -42,10 +41,8 @@ const DisplayRequests = ({ classes, user }) => {
                 <Link to={`/schedules/${schedule.id}`}>{schedule.name}</Link>
               </TableCell>
               <TableCell>{schedule.group.name}</TableCell>
-              {/* TODO: Make sure schedule object has this in the future */}
-              <TableCell>{schedule.type}</TableCell>
               <TableCell>
-                {schedule.startTime === numbers.distantPast ? '-' : moment.unix(schedule.startTime).format('LLL')}
+                {schedule.startTime === numbers.distantPast ? 'Ongoing' : moment.unix(schedule.startTime).format('LLL')}
               </TableCell>
               <TableCell>
                 {schedule.endTime === numbers.distantFuture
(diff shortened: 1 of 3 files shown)


web/src/screens/private/Dashboard.js
```diff
@@ -3,38 +3,120 @@ import PropTypes from 'prop-types';
...
+                <Link to={`/schedules/${schedule.id}`}>{schedule.name}</Link>
+              </TableCell>
+              <TableCell>{schedule.group.name}</TableCell>
+              {/* TODO: Make sure schedule object has this in the future */}
+              {/* <TableCell>{schedule.type}</TableCell> */}
+              <TableCell>
+                {schedule.startTime === 0 ? '-' : moment.unix(schedule.startTime).format('LLL')}
+              </TableCell>
+              <TableCell>
+                {schedule.endTime === 2147483647
```
sdunster: can we toss this in a `constants.js` or something? make one for `0` too if you're feeling game. Later on I hope we'll create a common lib for this kinda stuff (shared amongst web, client, server etc).

web/src/screens/private/Dashboard.js
```diff
@@ -3,38 +3,120 @@ import PropTypes from 'prop-types';
...
         </Typography>
       </Paper>
+      {/* TODO: Hide this when user has added their capabilities */}
+      <Paper className={classes.paper}>
+        <Typography component="p" align="center">
+          {/* TODO: update link t go edit profile page */}
+          You have no capabilities assigned for your user profile. <Link to="/groups">Click here to add capabilities</Link>
+        </Typography>
+      </Paper>
+      {displayRequests(classes, user)}
```
sdunster: just out of interest any reason you use a function here instead of defining a stateless component? I'm new to react.

justrossthings:
Screenshot:
image

web/src/screens/private/Dashboard.js
justrossthings: can do that but saw no justification if the component would not be used elsewhere

sdunster: You're 99% of the way there - with a few minor changes it becomes a component that is easier to test and reason about in the react view hierarchy. I'm not going to reject so you're free to merge without this if you still don't think it's worthwhile.
```
<DisplayRequests classes={classes} user={user} />
```

web/src/screens/private/Dashboard.js
```diff
@@ -3,38 +3,120 @@ import PropTypes from 'prop-types';
...
+import Toolbar from 'material-ui/Toolbar';
+import Table, { TableBody, TableCell, TableRow, TableHead } from 'material-ui/Table';
 
 import CURRENT_USER_QUERY from '../../graphql/current-user.query';
 
 import styles from './Dashboard.styles';
 
-const Dashboard = ({ classes, loading }) => {
+
+const displayRequests = (classes, user) => {
```
sdunster:
```
const DisplayRequests = ({ classes, user }) => {
```

web/src/screens/private/Dashboard.js
```diff
@@ -3,38 +3,120 @@ import PropTypes from 'prop-types';
...
+                  ? '-'
+                  : moment.unix(schedule.endTime).format('LLL')}
+              </TableCell>
+            </TableRow>
+          ))}
+        </TableBody>
+      </Table>
+    </Paper>
+  );
+};
```
sdunster:
```
DisplayRequests.propTypes({
  classes: PropTypes.shape({}).isRequired,
  <copy prop types for user from below>
});
```

web/src/components/DisplayRequests.js
```diff
@@ -0,0 +1,86 @@
...
+        // TODO: make sure this is returned in the future
+        // type: PropTypes.string.isRequired,
+        startTime: PropTypes.number.isRequired,
+        endTime: PropTypes.number.isRequired,
+      }),
+    ),
+  }),
+};
+
+export default DisplayRequests;
```
sdunster: please configure your editor to end files with newline.

web/src/constants.js
```diff
@@ -0,0 +1,4 @@
+export const numbers = {
+  undefined: 2147483647,
```
sdunster: sorry I should be more specific - it'd be great to use names like this: https://github.com/NSWSESMembers/availability-poc/blob/master/client/src/constants.js#L9

I think you should also avoid using `undefined` since it's a built in in JS.

web/src/constants.js
```diff
@@ -0,0 +1,4 @@
...
+  zero: 0,
+};
```
sdunster: newline

web/src/constants.js
sdunster: suggest `distantFuture` and `distantPast`

justrossthings:
Updated SS
image

PR #334, opened 2018-05-02
Push token registration framework

sdunster:
* `PushManager` to manage registration and notification event handling
* `<PushHandler>` to manage the lifecycle of push automatically as a React
  component
* Implement dummy push provider to prove it works.
* Successfully syncs dummy push tokens to the server even when there is an
  async update. Tested with multiple login, logout in one session.

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -11,6 +11,7 @@ import _ from 'lodash';
 import { RootNavigator } from './navigation';
 import auth from './state/auth.reducer';
 import local from './state/local.reducer';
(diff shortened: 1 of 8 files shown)


sjmcculloch:
lgtm - just need to resolve the app.js conflict.

sdunster:
untuck tests using Promises™

sdunster:
rebase

sdunster:
Fixed some bugs
* prevent registering more than once
* race condition during registration

PR #335, opened 2018-05-02
no bugsnag if in DEV

OSPFNeighbour:
use the __DEV__ flag to disable bugsnag.

I replaced the old check which I cannot understand how it was working as the logic was wrong as that always returns undefined because it has no typeof. I think I got lucky that the library protected its self from missing linked code.


OSPFNeighbour:
confirmed, doesn't shit the bed if BugsnagReactNative is not linked.

the react-native-bugsnag code protects its self

sdunster:
sweet merge it

PR #337, opened 2018-05-03
Add iOS push notification native module

sdunster:
* Enable running code in background in response to push notifications
* Confirmed we can succesfully get a push token from the native module in dev
  mode on my iPhone.


sdunster:
This is purely just adding the native module - there is no features implemented as a result of this. I'll commit this together with everything else once we've confirmed we can receive push notifications with it. I've proven it successfully gets push tokens.

sdunster:
Superseded by #343 - we'll review them together but land as separate commits.

sdunster:
oops ignore that

PR #339, opened 2018-05-03
[WIP] enable opt out for bug snag.

OSPFNeighbour:
new module for asking the question.

Ran into heaps of redux problems which my research lends me to believe is due to the reactJS debugger on android and the AsynStorage we use. I moved away from that bug still cant pull items from redux when using the debugger.

@sdunster can you link me to docs regarding why we use the immutable library for the redux store? I cant find any other examples on doing it that way, i rewrote them as per some redux docs before i worked out the debugger was my problem. Any thoughts on the way ive got it?

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -1,24 +1,30 @@
 import React from 'react';
-import { AsyncStorage, Alert } from 'react-native';
-import { Client as BugSnagClient } from 'bugsnag-react-native';
-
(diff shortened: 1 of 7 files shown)


sdunster:
ping 13 day old PR with WIP still in title, what's the plan?

OSPFNeighbour:
consensus was to just log peoples user name for now.

PR #341, opened 2018-05-04
Add group createdAt, updatedAt fields

sdunster:
* New `Date` type with code from:
  https://www.apollographql.com/docs/graphql-tools/scalars.html
* `Date` modified to return standard unix timestamp with second granularity.

screen shot 2018-05-03 at 22 49 43

### DIFF
--- a/server/src/resolvers.js
+++ b/server/src/resolvers.js
@@ -1,3 +1,6 @@
+import { GraphQLScalarType } from 'graphql';
+import { Kind } from 'graphql/language';
+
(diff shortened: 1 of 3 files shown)


sdunster:
* Add `createdAt` and `updatedAt` to get groups test.

OSPFNeighbour:
confirmed this doesnt hurt the client?

sdunster:
nope but I can't see any obvious reason why it would - the integration tests pass and they are a graphql client.

PR #343, opened 2018-05-04
APNS for iOS, new shell

sdunster:
Please see the individual commits for details.
I plan to land all of these by rebase onto master, once we've confirmed the following:
* iOS build on TestFlight is stable
* The JS code is safe on Android.


sdunster:
Confirmed this works with 0.1 (3) version of Callout installed from TestFlight using the production APNS environment.

img_2292

sdunster:
fix lint

PR #344, opened 2018-05-04
Push notifications for Android with Firebase/FCM

sdunster:
* Linked in firebase core, message, notification
* Disabled firebase_analytics & google ad id collection.
* Fixed up line indentation and formatting of Manifest.xml
* Add firebase push provider
 * Dynamic import - only try to import `react-native-firebase` if the underlying
  native module is available.

### DIFF
--- a/client/android/app/build.gradle
+++ b/client/android/app/build.gradle
@@ -102,7 +102,7 @@ android {
         applicationId "com.callout"
         minSdkVersion 16
         targetSdkVersion 22
-        versionCode 2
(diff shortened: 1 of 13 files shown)


sdunster:
Tested in simulator on iOS and on iPhone to make sure it doesn't break. It uploaded the dummy token and APNS token but no mention of FCM.

sdunster:
Ok this is just blocked by me building a new apk and testing that all works. Once the new apk is verified good we'll merge this.

PR #350, opened 2018-05-05
Save device name to the server

sdunster:
* Allows us to more easily track which device is doing what during testing
* Device name comes from `react-native-device-info`.
* Synced to server whenever we update the push token.
* Re-work the `<PushHandler>` tests to use Jest Mock Functions.


sdunster:
btw this has a backwards incompatible server change so devices will not successfully submit push tokens until they update. I don't think it will crash the client.

PR #352, opened 2018-05-05
Upgrade all server deps

sdunster:


sdunster:
I did some basic testing locally - the server booted and seemed to accept mutations. Lets see if travis finds anything.

PR #355, opened 2018-05-06
Separate out components from screens and add tests

sdunster:
* auth/Home
* auth/SignIn
* auth/SignUp
* burger/Root
* burger/Profile

### DIFF
--- a/client/src/screens/auth/components/SignIn.js
+++ b/client/src/screens/auth/components/SignIn.js
@@ -1,0 +1,77 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { View, ActivityIndicator } from 'react-native';
+
+import { Container, Holder } from '../../../components/Container';
+import { Input } from '../../../components/Input';
+import { Button } from '../../../components/Button';
+
+class SignIn extends Component {
+  state = {
+    username: '',
+    password: '',
+  }
+
+  onChangeUsername = (value) => {
+    this.setState({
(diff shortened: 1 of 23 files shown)


client/src/screens/auth/components/SignIn.js
```diff
@@ -0,0 +1,77 @@
...
+          <Input
+            type="password"
+            placeholder="Password"
+            value={password}
+            onChangeText={this.onChangePassword}
+          />
+          <Button text="Sign In" onPress={this.onSignIn} disabled={loading} />
+          {loading && (
+            <View
+              style={{
```
OSPFNeighbour: We have a component to do what this view is doing. Think it's called Container Middle or something like that.

client/src/screens/auth/components/SignUp.js
```diff
@@ -0,0 +1,90 @@
...
+          <Input
+            type="email"
+            placeholder="Email"
+            value={email}
+            onChangeText={this.onChangeEmail}
+          />
+          <Button text="Register" onPress={this.onSignUp} disabled={loading} />
+        </Holder>
+        {loading && (
+          <View
```
OSPFNeighbour: Use existing container as per other comment

client/src/screens/auth/components/SignIn.js
sdunster: yeah got a follow up PR for that

PR #356, opened 2018-05-06
Test push mutation

sdunster:
* In burger menu you can tap 'Send test push notification' and it will hit the
  server via mutation and have it send a push notification to the current
  device.
* Built a push service abstraction so you can call one function and it will
  automatically send via the appropriate service for the device.
* Currently only APNS is supported.

### DIFF
--- a/client/src/graphql/send-test-push.mutation.js
+++ b/client/src/graphql/send-test-push.mutation.js
@@ -1,0 +1,7 @@
+import gql from 'graphql-tag';
+
+export default gql`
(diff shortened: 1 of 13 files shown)


sdunster:
Yeah I was thinking about adding that, perhaps you could do it in a follow up if interested.

sdunster:
I tricked my simulator into sending my real iPhone's push token to my dev server. This was the result on my real iPhone.
meh

PR #357, opened 2018-05-06
Use <Progress> component instead of custom stuff for auth screens

sdunster:
* It's not as pretty as before but it's better than custom stuff for now.


sdunster:
Screenshot is hard. It displays an activity indicator as expected I promise.

PR #359, opened 2018-05-06
Always display version number

sdunster:
* This might break on shells with no device info but we will address this in future.
* The existing device info check didn't work.
* Display version as `<version> (<build>)` ie `0.1 (6)`
* Move function to `deviceInfo` selector.
* Rewrite callers.


sdunster:
screen shot 2018-05-06 at 10 50 14

PR #368, opened 2018-05-07
Fix for device info on android

OSPFNeighbour:
version++ to support included options


sdunster:
#375 includes this and more

PR #370, opened 2018-05-08
Fixes for push manager on client

OSPFNeighbour:
Fixed dummy undefined error on init when not logged in

Fixed FCM remove error when logging out

### DIFF
--- a/client/src/push/manager.js
+++ b/client/src/push/manager.js
@@ -21,7 +21,10 @@ class PushManager {
   async init() {
     const promises = [];
 
-    this.services.dummy = await initDummy(this);
+    const dummy = await initDummy(this);
+    if (dummy !== null) {
+      this.services.dummy = dummy;
+    }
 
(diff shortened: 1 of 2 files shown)


client/src/push/manager.js
```diff
@@ -21,7 +21,10 @@ class PushManager {
   async init() {
     const promises = [];
 
-    this.services.dummy = await initDummy(this);
+    const dummy = await initDummy(this);
```
sdunster: I'm not convinced this fixes it because I don't see how the thing that we call can ever return null. I think we should do more investigation.

sdunster: well let me clarify. It probably does fix the bug as we see it but it's masking something else. This same situation can probably happen for the other providers.

OSPFNeighbour: Yea its protecting the problem not fixing it.

PR #372, opened 2018-05-08
Web Event Management

sjmcculloch:
- Added createEvent & updateEvent methods to server logic
- Used fixture data for locations to allow quick entry
- Updated event routing
- Added step captioning to allow previous step entries to be visible when not selected

image

image

image


server/src/logic.js
```diff
@@ -335,6 +335,99 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
...
+              return Promise.reject(Error('Invalid group!'));
+            }
+            return Creators.event({
+              name,
+              details,
+              sourceIdentifier,
+              permalink,
+              group,
+            }).then((event) => {
+              if (eventLocations === undefined) return;
```
sdunster: have you verified this works? How is this different to `typeof eventLocations === 'undefined'` which I think was the recommended way to do this.

web/src/screens/private/Events/EditEvent.js
```diff
@@ -4,31 +4,159 @@ import { compose } from 'recompose';
...
+  }
+
+  onGroupChange = (e) => {
+    const groupId = e.target.value;
+    let name = '';
+
+    const group = this.props.user.groups.find(g => g.id === parseInt(groupId, 10));
+    if (group !== undefined) ({ name } = group);
+
+    this.setState(() => ({ groupId, groupName: name }));
```
sdunster: any reason you use the callback form of setState() instead of passing a dict directly? might be good to capture this in comments

web/src/screens/private/Events/EditEvent.js
```diff
@@ -131,8 +453,46 @@ const userQuery = graphql(CURRENT_USER_QUERY, {
...
 });
 
-export default compose(connect(mapStateToProps), withStyles(styles), userQuery)(EditEvent);
+export default compose(
+  connect(mapStateToProps),
+  withStyles(styles),
+  createEvent,
+  updateEvent,
+  userQuery,
+)(EditEvent);
```
sdunster: this component is massive - in future I think we should try to break it up. not now :)

sdunster:
needs manual rebase

web/src/screens/private/Events/EditEvent.js
sjmcculloch: Yes there are definitely parts that can be moved to components.

web/src/screens/private/Events/EditEvent.js
sjmcculloch: Oversight - I'll correct.

server/src/logic.js
sjmcculloch: I'll update it to preferred method

PR #383, opened 2018-05-09
Update to the SES hook to create jobs from a HTTP POST.

OSPFNeighbour:
Updated the hook code to create jobs.
Update the model and schema to support primaryLocation.
Update the location creator to handle primaryLocation flag.

We might want to change the model to work backwards to how i did it. I hold the ID of the primary group but we don't currently reuse locations so we could just have a boolean in the location that makes it the default.

I had to do some fuckery with the second one-to-one in the DB with the foreign key that I don't fully understand. Something @sdunster might want to check.

Made a bunch of assumptions in this code and more than happy for people to challenge them.

- Im looking for a group that has the LHQ name that the job is assigned to in it, if nothing found it will use group 1.
- Im making LHQ the primary location for the job.
- Im making the details of the job the situation on scene plus a join of the tags array.

if you run the python script in /scripts/hook.py you should get the following event created.

```
{
          "name": "Storm - 12 GILMORE STREET, WEST WOLLONGONG",
          "details": "Things about stuff [#Flooded]",
          "primaryLocation": {
            "name": "hlq",
            "detail": "WOL"
          },
          "eventLocations": [
            {
              "name": "hlq",
              "detail": "WOL",
              "locationLatitude": -34.40247,
              "locationLongitude": 150.89495
            },
            {
              "name": "scene",
              "detail": "12 GILMORE STREET, WEST WOLLONGONG",
              "locationLatitude": -34.41949,
              "locationLongitude": 150.87497
            }
          ]
        }
      ]
    }
```

### DIFF
--- a/server/src/callback.js
+++ b/server/src/callback.js
@@ -1,25 +1,52 @@
+import { Op } from 'sequelize';
+
+
 const SESCallback = {
   creator: undefined,
+  models: undefined,
   eventCallback(req, res) {
     const job = {
-      name: req.body.JobType.Name,
+      name: `${req.body.JobType.Name} - ${req.body.Address.PrettyAddress}`,
       identifier: req.body.Identifier,
       description: req.body.JobType.Description,
+      SituationOnScene: req.body.SituationOnScene,
+      tags: req.body.Tags,
     };
+    // look for a group with the LHQ name in it, or use group id:1
+    this.models.Group.findOne({ where: { name: { [Op.like]: `%${req.body.EntityAssignedTo.Code}%` } } }).then((group) => {
+      Promise.resolve(this.creator.event({
+        name: job.name,
+        details: `${job.SituationOnScene} [${job.tags.map(g => `#${g.Name}`).join(', ')}]`,
+        identifier: job.identifier,
+        permalink: `https://beacon.com/job/${job.identifier}`,
+        group: group || { id: 1 },
+      }).then((event) => {
+        Promise.all(
+          [
+            this.creator.eventLocation({
+              name: 'hlq',
+              detail: req.body.EntityAssignedTo.Code,
+              icon: 'scene',
+              locationLatitude: req.body.EntityAssignedTo.Latitude,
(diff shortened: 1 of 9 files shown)


server/src/callback.js
```diff
@@ -1,36 +1,82 @@
...
+      Promise.resolve(this.creator.event({
+        name: job.name,
+        details: `${job.SituationOnScene} [${job.tags.map(g => `#${g.Name}`).join(', ')}]`,
+        identifier: job.identifier,
+        permalink: `https://beacon.com/job/${job.identifier}`,
+        group: group || { id: 1 },
+      }).then(event => Promise.all(
+        [
+          this.creator.eventLocation({
+            name: 'hlq',
```
sdunster: lhq?

OSPFNeighbour:
Error handling is not right, but ill fix that later

server/src/callback.js
```diff
@@ -1,36 +1,82 @@
...
+            event,
+          }),
+        ],
+      ).then(() => {
+        // we are done?
+        const result = {
+          event,
+        };
+        res.send(JSON.stringify(result));
+      }).catch((err) => {
```
sdunster: this part is impossible to read. which catch is for which part? Indentation would help but I think some comments are necessary too.

server/src/logic.js
```diff
@@ -335,6 +336,12 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
       eventLocations(event) {
         return event.getEventlocations();
       },
+      primaryLocation(event) {
+        // to avoid a cyclic table associations we are calling 'constraints: false'
```
sdunster: maybe mention this is in the model definition:
```
// to avoid a cyclic table association we create the EventModel with `constraints: false`
```

OSPFNeighbour: already mentioned.

server/src/callback.js
OSPFNeighbour: nice catch

OSPFNeighbour:
Land #385 first

PR #385, opened 2018-05-09
fixed errors found from blank data

OSPFNeighbour:
Found a few places that really dont like blank or invalid data.


OSPFNeighbour:
Land this before #383

PR #386, opened 2018-05-09
Suppres isMounted deprecation warning as it is a false-positive

sdunster:
* Proper fix coming from react-native soon:
  * https://github.com/facebook/react-native/issues/18868#issuecomment-387627007


OSPFNeighbour (approved):
cheater :P

gaearon:
Should be fixed in 0.57.

PR #394, opened 2018-05-09
Event Map Icon Rewrite

OSPFNeighbour:
Removed the shitty custom images for map icons. Now using native Icon component and passing icon names directly from the event.
Added new components for IconMarker and styles to make them purdy.

screenshot_20180509-174652


client/src/components/Icon/index.js
```diff
@@ -12,14 +13,15 @@ const Icon = function iconBuilder(props) {
...
+    case 'mi': return (<MaterialIcons {...props} name={iconName} />);
     case 'fnd': return (<Foundation {...props} name={iconName} />);
     case 'oi': return (<Octicons {...props} name={iconName} />);
 
     default: return (<FontAwesome {...props} name={props.name} />);
   }
 };
 Icon.propTypes = {
-  name: PropTypes.string,
+  props: PropTypes.string,
```
sdunster: no pretty sure this should be name. The other props should be listed too but eslint isn't smart enough to require you to add them. It's fine to leave them out imo.

sdunster:
lgtm
screen shot 2018-05-09 at 22 38 01

client/src/screens/events/Detail.js
```diff
@@ -274,12 +273,13 @@ class Detail extends Component {
         ))}
             {mapEventLocations.map(marker => (
               <Marker
```
sjmcculloch: Missing "key" attribute - you'll get "Warning: Each child in an array..."

sdunster: doh I noticed that too but assumed it wasn't related :/

PR #408, opened 2018-05-10
GPS error change on Event Details

OSPFNeighbour:
Old one was throwing false positives.

Moved to a timer so that any GPS source will be checked, not just the fuzzy or accurate one.

### DIFF
--- a/client/src/screens/events/Detail.js
+++ b/client/src/screens/events/Detail.js
@@ -1,7 +1,7 @@
 /* global navigator */
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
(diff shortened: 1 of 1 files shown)


client/src/screens/events/Detail.js
```diff
@@ -207,17 +218,18 @@ mapZoomMe = () => {
...
+        console.log(err);
       });
 
       // start a accurate fix watcher
       this.geoWatch = navigator.geolocation.watchPosition(
         (position) => {
           this.processReturnedLocation(position, true);
+        }, (err) => {
+          console.log(err);
         },
```
sdunster: Alerts instead of console log?

OSPFNeighbour: if I take that out I get a RedBox console.error popup. no idea what its doing it or how to stop it.

PR #409, opened 2018-05-10
Ignore .env

sdunster:
* makes it safer to store APNS_KEY there


OSPFNeighbour:
Nice

sdunster:
not even sure this even works but it can't hurt to ignore it to prevent one of us actually committing a secret

PR #410, opened 2018-05-10
Better logging/debug for push on server side

sdunster:


server/src/logic.js
```diff
@@ -616,7 +616,7 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
       async sendTestPush(ctx) {
         console.log(ctx);
         const device = await getAuthenticatedDevice(ctx);
-        const result = push.sendPush({
+        const result = await push.sendPush({
```
OSPFNeighbour: For my own sake, can you explain this change to me>

sdunster: it was a typo. `push.sendPush()` returns a promise which means that `result` was a promise. We would continue past this line even though the promise hadn't been fulfilled yet (ie we're still in the process of sending the push).
With await we don't proceed beyond that line until the promise is fulfilled and when it is we assign the result to `result`.
In this situation it's functionally the same because GraphQL awaits all promises before returning to the client but the original intention when I wrote this is pretty clearly to await the promise because I called the value `result`, rather than `promise`. `result` implies that `sendPush` is done and we have a result telling us whether it succeeded or not.

PR #417, opened 2018-05-10
Modifications to mutations and broker to send delayed push messages

### LINKED ISSUE #412: Make a delayed option for the sendTestPush mutation
if you call it with `delayed=true` then it waits 5 seconds before sending so you have time to exit the app.

OSPFNeighbour:
Closes #412 

freebee util to make async wait

### DIFF
--- a/client/src/screens/burger/Root.js
+++ b/client/src/screens/burger/Root.js
@@ -124,17 +124,17 @@ class BurgerScreen extends Component {
     }
   };
 
-TestPushPrompt = () => {
-  Alert.alert(
-    'When?',
-    'Instantly or in 5 seconds?',
-    [
-      { text: 'Instantly', onPress: () => this.sendTestPush(false) },
-      { text: 'In 5 Seconds', onPress: () => this.sendTestPush(true) },
-    ],
-    { cancelable: false },
-  );
-}
+  testPushPrompt = () => {
+    Alert.alert(
+      'When?',
+      'Instantly or in 5 seconds?',
+      [
+        { text: 'Instantly', onPress: () => this.sendTestPush(false) },
+        { text: 'In 5 Seconds', onPress: () => this.sendTestPush(true) },
+      ],
+      { cancelable: false },
+    );
+  }
 
   sendTestPush = (delay) => {
     this.props.sendTestPush(delay).catch((e) => {
@@ -153,7 +153,7 @@ TestPushPrompt = () => {
         onCheckForUpdate={this.checkForUpdate}
         onSubmitFeedback={this.handleSubmitFeedback}
         onTestBugsnag={this.crashReport}
-        onSendTestPush={this.TestPushPrompt}
+        onSendTestPush={this.testPushPrompt}
         onShowParams={this.showParams}
         onLogout={this.logout}
       />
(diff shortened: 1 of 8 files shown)


client/src/screens/burger/Root.js
```diff
@@ -121,8 +124,20 @@ class BurgerScreen extends Component {
     }
   };
 
-  sendTestPush = () => {
-    this.props.sendTestPush().catch((e) => {
+TestPushPrompt = () => {
```
sdunster: this is another example of where eslint is failing to warn about indentation. Also please don't start a regular instance method with a capital letter.

server/src/push/broker.js
```diff
@@ -30,15 +31,21 @@ const sendPush = ({ devices, message }) => {
...
     }
-
     return Promise.all(promises).then(() => true);
   });
 };
 
+const sendTestPush = async ({ devices, message, delay }) => {
+  if (delay) {
+    await sleep(5000);
+    return sendPush({ devices, message });
```
sdunster: delete this line

server/src/push/broker.js
```diff
@@ -30,15 +31,21 @@ const sendPush = ({ devices, message }) => {
...
+  }
+  return sendPush({ devices, message });
```
sdunster:
```
return await sendPush({ devices, message });
```

server/src/schema.js
```diff
@@ -141,6 +141,10 @@ export const Schema = [
     groupId: Int!
   }
 
+  input sendTestPushInput {
```
sdunster: why does this start with lower case when every other input starts with upper case?

server/src/push/broker.js
```diff
@@ -1,5 +1,6 @@
 import { sendPush as apns } from './apns';
 import { sendPush as fcm } from './fcm';
+import sleep from '../utils';
```
sdunster:
```
import { sleep } from '../utils';
```

server/src/utils.js
```diff
@@ -0,0 +1,6 @@
+const sleep = time => new Promise((resolve) => {
+  setTimeout(() => resolve(), time);
+});
+
+
+export default sleep;
```
sdunster: this will definitely have more things in it at some point:
```
export default { sleep };
```
import with:
```
import { sleep } from './utils';
```

OSPFNeighbour: results in undefined function errors

server/src/push/broker.js
sdunster: bleh.. fine. Just don't forget to add it if you add it if you remove the return and add some additional code afterwards.

server/src/utils.js
sdunster: oh I know why.. you'd need to import it as:
```
import utils from '../utils';
```
and use:
```
await utils.sleep(1000);
```
but that's not good for utils so this is fine for now.

server/src/push/broker.js
```diff
@@ -30,15 +31,20 @@ const sendPush = ({ devices, message }) => {
...
 };
 
+const sendTestPush = async ({ devices, message, delay }) => {
+  if (delay) {
+    await sleep(5000);
+  }
+  return sendPush({ devices, message });
+};
+
 // eslint-disable-next-line import/prefer-default-export
```
sdunster: you can remove this now

PR #419, opened 2018-05-11
Fixed GPS location not updating. Added Halo circle to show location accuracy

### LINKED ISSUE #418: GPS Marker not moving
GPS user marker not moving on the map.

OSPFNeighbour:
Fixed #418 
Fixed missing keys on LocationMarker

Added a circle around the user marker to show location accuracy.

Added tests for new component.

screenshot 2018-05-11 11 24 27

### DIFF
--- a/client/src/components/MapMarker/AccuracyHalo.js
+++ b/client/src/components/MapMarker/AccuracyHalo.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Circle } from 'react-native-maps';
+
+const AccuracyHalo = ({ myPosition }) => {
+  if (
+    !myPosition ||
+    !myPosition.latitude ||
+    !myPosition.longitude ||
+    !myPosition.accuracy
+  ) return null;
+
+
+  return (
+    <Circle
+      key="myPositionHalo"
+      center={myPosition}
+      radius={myPosition.accuracy}
+      strokeColor="transparent"
+      fillColor="rgba(135,206,250,0.3)"
+    />
+  );
+};
+
+AccuracyHalo.propTypes = {
+  myPosition: PropTypes.shape({
+    latitude: PropTypes.number,
+    longitude: PropTypes.number,
+    heading: PropTypes.number,
+    accuracy: PropTypes.number,
+  }),
+};
+
+export default AccuracyHalo;
--- a/client/src/components/MapMarker/MyLocationMarker.js
+++ b/client/src/components/MapMarker/MyLocationMarker.js
@@ -8,7 +8,6 @@ import { Marker } from 'react-native-maps';
 
 import styles from './styles';
 
-
 const ANCHOR = { x: 0.5, y: 0.5 };
 
 const MyLocationMarker = ({ myPosition }) => {
@@ -19,6 +18,7 @@ const MyLocationMarker = ({ myPosition }) => {
 
   return (
     <Marker
+      key={(myPosition.longitude + myPosition.latitude).toString()}
       anchor={ANCHOR}
       style={styles.mapMarker}
       coordinate={myPosition}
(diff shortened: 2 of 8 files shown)


client/src/components/MapMarker/MyLocationMarker.js
```diff
@@ -19,6 +18,7 @@ const MyLocationMarker = ({ myPosition }) => {
 
   return (
     <Marker
+      key={(myPosition.longitude + myPosition.latitude).toString()}
```
sdunster: I don't think this should change as the marker moves. I suggest using a static string like "my-location" or even add the user's ID number or something.

client/src/components/MapMarker/AccuracyHalo.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Circle } from 'react-native-maps';
+
+const AccuracyHalo = ({ myPosition }) => {
+  if
```
sdunster:
```
if (
  !myPosition ||
  !myPosition.latitude ||
  !myPosition.longitude ||
  !myPosition.accuracy
) return null;
```

client/src/components/MapMarker/__tests__/__snapshots__/AccuracyHalo.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders correctly 1`] = `null`;
```
sdunster: check your logic - both of these render to null which feels like a bug

client/src/components/MapMarker/MyLocationMarker.js
OSPFNeighbour: This was a work around for a 'bug' in maps, but its looks like its fixed. will remove.

client/src/components/MapMarker/__tests__/__snapshots__/AccuracyHalo.js.snap
OSPFNeighbour: My spelling is just that bad

client/src/components/MapMarker/AccuracyHalo.js
OSPFNeighbour: was wonder how to do that

client/src/components/MapMarker/__tests__/__snapshots__/AccuracyHalo.js.snap
sdunster: I dont see it

client/src/screens/events/Detail.js
```diff
@@ -238,7 +238,7 @@ mapZoomMe = () => {
     const myLastPosition = this.state.myPosition;
     const myPosition = position.coords;
     // if we already have a high accuracy location ignore this fuzzy location
-    if (!HighAccuracy && !myLastPosition) {
+    if ((!HighAccuracy && !myLastPosition) || HighAccuracy) {
```
sdunster: `highAccuracy` - change the function param to lower case

client/src/screens/events/Detail.js
```diff
@@ -214,7 +214,7 @@ mapZoomMe = () => {
       return;
     }
     if (this.geoWatch == null) { // If this is the first run
-      // start a single fuzzy location fix aqusition
+    // start a single fuzzy location fix aqusition
```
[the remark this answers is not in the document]
sdunster: maybe cause you also have a comment not he same line as the if statement? move that one before the `if`. Also while you are there `s/aqusition/acquisition` (i also had to consult dictionary)

client/src/components/MapMarker/__tests__/AccuracyHalo.js
```diff
@@ -6,8 +6,7 @@ import AccuracyHalo from '../AccuracyHalo';
 const valid = {
   latitude: 1,
   longitude: 1,
-  heading: 1,
-  accurary: 1,
+  accuracy: 1,
```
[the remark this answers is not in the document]
sdunster: fair point, but it's good to write all your code twice so you can be really sure of every decision/keystroke :P

PR #422, opened 2018-05-12
Apollo Client V2

### LINKED ISSUE #146: Reproducible GraphQL endpoints not loading on client side
The client reproducibly doesn't load graphQL endpoints when switching accounts after killing the app while being logged in.

iOS Repro:
* log into app
* exit app and kill
* open app
* logout
* login
* events now spins forever

This problem doesn't occur if you logout, exit and then kill the app.

OSPFNeighbour:
* Upgrade apollo client libs
* Rewrite middleware to support new `apollo-client` APIs.
* Handle unauthorised errors by parsing the GraphQL errors rather than HTTP error codes because Apollo doesn't always expose them anymore: https://github.com/apollographql/apollo-link/issues/300
* Closes #146

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -13,7 +13,11 @@
     "lint": "eslint src"
   },
   "dependencies": {
-    "apollo-client": "^1.9.3",
+    "apollo-cache-inmemory": "^0.2.0-rc.4",
+    "apollo-client": "^2.0.0-rc.7",
+    "apollo-link": "^1.2.2",
+    "apollo-link-error": "^1.0.9",
+    "apollo-link-http": "^1.5.4",
     "babel-core": "^6.26.0",
     "buffer": "^5.0.7",
     "bugsnag-react-native": "~2.9.1",
--- a/client/package.json
+++ b/client/package.json
@@ -13,8 +13,8 @@
     "lint": "eslint src"
   },
   "dependencies": {
-    "apollo-cache-inmemory": "^0.2.0-rc.4",
-    "apollo-client": "2.2.3",
+    "apollo-cache-inmemory": "^1.2.1",
+    "apollo-client": "2.3.1",
     "apollo-link": "^1.2.2",
     "apollo-link-error": "^1.0.9",
     "apollo-link-http": "^1.5.4",
--- a/client/package.json
+++ b/client/package.json
@@ -14,7 +14,7 @@
   },
   "dependencies": {
     "apollo-cache-inmemory": "^1.2.1",
-    "apollo-client": "2.3.1",
+    "apollo-client": "^2.3.1",
     "apollo-link": "^1.2.2",
     "apollo-link-error": "^1.0.9",
     "apollo-link-http": "^1.5.4",
--- a/client/package.json
+++ b/client/package.json
@@ -14,7 +14,7 @@
   },
   "dependencies": {
     "apollo-cache-inmemory": "^0.2.0-rc.4",
-    "apollo-client": "^2.0.0-rc.7",
+    "apollo-client": "2.2.3",
     "apollo-link": "^1.2.2",
     "apollo-link-error": "^1.0.9",
     "apollo-link-http": "^1.5.4",
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -3,7 +3,7 @@ import { AsyncStorage, Alert } from 'react-native';
 import { Client as BugSnagClient } from 'bugsnag-react-native';
 import ApolloClient from 'apollo-client';
 import { HttpLink } from 'apollo-link-http';
-import { ApolloLink, concat } from 'apollo-link';
+import { ApolloLink } from 'apollo-link';
 
 import { InMemoryCache } from 'apollo-cache-inmemory';
 import { onError } from 'apollo-link-error';
@@ -39,11 +39,13 @@ const httpLink = new HttpLink({ uri: GRAPHQL_ENDPOINT });
 // replaces networkInterface.use(applyMiddleware...)
 const authMiddleware = new ApolloLink((operation, forward) => {
   // add the authorization to the headers
-  operation.setContext({
-    headers: {
-      authorization: store.getState().auth || null,
-    },
-  });
+  if (store.getState().auth.token) {
+    operation.setContext({
+      headers: {
+        authorization: store.getState().auth.token ? `Bearer ${store.getState().auth.token}` : null,
+      },
+    });
+  }
 
   return forward(operation);
 });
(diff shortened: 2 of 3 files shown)


client/package.json
```diff
@@ -13,7 +13,11 @@
     "lint": "eslint src"
   },
   "dependencies": {
-    "apollo-client": "^1.9.3",
+    "apollo-cache-inmemory": "^0.2.0-rc.4",
+    "apollo-client": "^2.0.0-rc.7",
```
sjmcculloch: The web app is using "apollo-client": "^2.2.3" - any reason you are using an older version?

client/src/app.js
```diff
@@ -24,28 +28,56 @@ if (!__DEV__) {
...
-      }
-      next();
+const httpLink = new HttpLink({ uri: GRAPHQL_ENDPOINT });
+
+// replaces networkInterface.use(applyMiddleware...)
+const authMiddleware = new ApolloLink((operation, forward) => {
+  // add the authorization to the headers
+  operation.setContext({
+    headers: {
+      authorization: store.getState().auth || null,
```
sjmcculloch: We should align the web with the mobile app (see below with what is working for web) - I'm open to using either:

```
const authMiddleware = new ApolloLink((operation, forward) => {
  if (store.getState().auth.token) {
    operation.setContext({
      headers: {
        authorization: store.getState().auth.token ? `Bearer ${store.getState().auth.token}` : null,
      },
    });
  }

  return forward(operation);
});
```

client/package.json
OSPFNeighbour: guide I followed used this version

OSPFNeighbour:
I was hoping to capture all the network error alerts in here with error handler but they still throw as unhanded, even though it does do a console.log. thoughts?

sdunster:
this is still titled wip

sdunster:
I hit problems with this. It seems the new stuff doesn't expose the HTTP error code:
screen shot 2018-05-16 at 12 36 43

basically, `networkError` is undefined at L95 which causes an exception. Can you try to repro? In this case it happened because my server doesn't have the device record my simulator is using which means that GraphQL returns a HTTP 401 with the error `"Unauthorized: Device not found"`. Normally this would trigger `warnLogout()` and the actual logout on L96-97.

You're definitely running the same versions right? Here is my `yarn list | grep apollo`:
```
├─ apollo-cache-inmemory@0.2.0-rc.4
│  ├─ apollo-cache@^0.2.0-rc.3
│  ├─ apollo-utilities@^0.2.0-rc.3
│  ├─ apollo-utilities@0.2.0-rc.3
├─ apollo-cache@0.2.0-rc.3
│  ├─ apollo-utilities@^0.2.0-rc.3
│  └─ apollo-utilities@0.2.0-rc.3
├─ apollo-client@2.2.3
│  ├─ apollo-cache@^1.1.2
│  ├─ apollo-cache@1.1.8
│  │  └─ apollo-utilities@^1.0.12
│  ├─ apollo-link-dedup@^1.0.0
│  ├─ apollo-link@^1.0.0
│  ├─ apollo-utilities@^1.0.6
├─ apollo-link-core@0.5.4
├─ apollo-link-dedup@1.0.9
│  └─ apollo-link@^1.2.2
├─ apollo-link-error@1.0.9
│  └─ apollo-link@^1.2.2
├─ apollo-link-http-common@0.2.4
│  └─ apollo-link@^1.2.2
├─ apollo-link-http@1.5.4
│  ├─ apollo-link-http-common@^0.2.4
│  └─ apollo-link@^1.2.2
├─ apollo-link@1.2.2
│  ├─ apollo-utilities@^1.0.0
├─ apollo-utilities@1.0.12
│  └─ apollo-utilities@^1.0.12
├─ react-apollo@1.4.16
│  ├─ apollo-client@^1.4.0
│  ├─ apollo-client@1.9.3
│  │  ├─ apollo-link-core@^0.5.0
```

Sidenote: there's still lots of rc versions there is that intentional? `package.json` implies it is: `"apollo-cache-inmemory": "^0.2.0-rc.4"`

OSPFNeighbour:
Unsure why the problem that Sam found didnt show up in my multiple labs. im guessing for some reason I didnt trigger a 401 even though I had blown away the sql database.

Ive updated the dependencies away from the RC's that Sam pointed out.

Ive put in a work around for the code that catches 401's. Theres several logged Issues for the problem against apollo-link-error so I would expect a fix pretty soon. Interim we can work with what ive got.

sdunster:
I request clarification of your last comment. Were you able to successfully reproduce this in the end?

client/src/app.js
```diff
@@ -57,45 +91,28 @@ const warnLogout = _.debounce(() => {
...
-              warnLogout();
-              store.dispatch(logout());
-            }
-            next();
-          });
-      }
-    },
-  },
-]);
+// Hack to get around known issue with networkError not having network header errors
```
sdunster: 👍

client/src/app.js
```diff
@@ -57,45 +91,28 @@ const warnLogout = _.debounce(() => {
...
+// https://github.com/apollographql/apollo-link/issues/300
+const logoutLink = onError(({ graphQLErrors }) => {
+  if (graphQLErrors && graphQLErrors[0].message.includes('Unauthorized')) {
```
sdunster: yeah this is not ideal but it'll work for now. We should do some research into how others pass these kinds of errors down and how they parse the result to look for specific things like this. It's weird to me that GraphQL doesn't provide a way to pass specific error codes or something.

client/src/app.js
```diff
@@ -57,45 +91,28 @@ const warnLogout = _.debounce(() => {
...
+    store.dispatch(logout());
+    return warnLogout();
+  }
+  return null;
+});
+
+// This should work once apollo-link issue #300 is resolved
```
sdunster: I really don't like leaving commented code in the repo - stash it in a branch or something. I won't reject the PR over this.

client/package.json
```diff
@@ -13,7 +13,11 @@
     "lint": "eslint src"
   },
   "dependencies": {
-    "apollo-client": "^1.9.3",
+    "apollo-cache-inmemory": "^1.2.1",
+    "apollo-client": "2.3.1",
```
sdunster: you should probably add the caret here so we can do minor upgrades unless there's a particular reason you want to force `2.3.1` only.
```
"apollo-client": "^2.3.1",
```

sdunster:
I updated your PR description - please make sure you use it when you merge.

client/src/app.js
OSPFNeighbour: Apollo does... its just broken. The preferred fix for this issue was with a typecast but it didnt work for me.

PR #429, opened 2018-05-14
Groups view

justrossthings:
view user's joined groups
browse all groups
filters/search for both


web/src/screens/private/Groups/ViewGroups.js
```diff
@@ -3,25 +3,173 @@ import PropTypes from 'prop-types';
...
+    const { classes, loading, user } = this.props;
+    const { tab, locationFilter, capabilityFilter, searchFilter } = this.state;
+
+    if (loading) {
+      return <CircularProgress className={classes.progress} size={50} />;
+    }
+
+    const allGroups = user.organisation.groups;
+    const myGroups = user.groups;
+    const allLocations = _.keys(_.transform(allGroups, (result, group) => {
```
sdunster: this logic is pretty hard to follow - it'd be good to toss a comment in explaining the result for future readers
```
// get all tags of type `orgStructure` from all groups in the organisation
```
(even better would be to make a function that does this)

web/src/screens/private/Groups/ViewGroups.js
```diff
@@ -3,25 +3,173 @@ import PropTypes from 'prop-types';
...
+      _.forEach(group.tags, (tag) => {
+        if (tag.type === 'orgStructure') {
+          result[tag.name] = true;
+        }
+      });
+    }, {}));
+    const allCapabilities = _.keys(_.transform(allGroups, (result, group) => {
```
sdunster: as above

justrossthings:
Screenshot:

pasted_image_5_15_18__10_14_am

web/src/graphql/current-user.query.js
```diff
@@ -8,11 +8,28 @@ export default gql`
       displayName
       email
       username
+      organisation {
+        groups {
```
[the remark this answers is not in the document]
sjmcculloch: I mentioned this also @sdunster - we'll have to do a bit of an audit on who uses what for the refetches.

sjmcculloch:
I can help with that styling @justrossthings once it's merged.

PR #447, opened 2018-05-20
Prevent events map from being re-rendered every 5s

### LINKED ISSUE #446: Event map redraws when data changes
Full map is redrawn when a marker moves on the event map causing the map to zoom to home coordinates then back to bounds of markers.

to reproduce: 

have an event open with a moving tracked user.

### LINKED ISSUE #466: Job Acknowledgment - Event Response
When responding to immediate / rescue jobs an acknowledge job should be implemented for Unit Controllers to accept jobs

OSPFNeighbour:
Rewrote previous componentWillReceiveProps function into getDerivedStateFromProps.

Minor timer name changes and white space changes.

I was incorrectly returning from the function which was causing the map to get redrawn every 5 seconds.

Fixes #446

### DIFF
--- a/client/src/screens/events/Detail.js
+++ b/client/src/screens/events/Detail.js
@@ -71,6 +71,24 @@ class Detail extends Component {
     return mapMarkers;
   }
 
+  static getDerivedStateFromProps(newProps) {
+    // catch incoming props and generate the marker states
+    const { event, loading, auth } = newProps;
+    let responseMarkers = {};
+    let eventMarkers = {};
+    if (!loading && event) {
+      responseMarkers = Detail.makeResponseMarkers(
+        auth.id, event.responses,
+      );
+      eventMarkers = Detail.makeEventLocations(event.eventLocations);
+      return {
+        responseMarkers,
+        eventMarkers,
+      };
+    }
+    return null;
+  }
+
   state = {
     myPosition: null,
     eventMarkers: null,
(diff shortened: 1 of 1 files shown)


client/src/screens/events/Detail.js
```diff
@@ -71,6 +71,24 @@ class Detail extends Component {
     return mapMarkers;
   }
 
+  static getDerivedStateFromProps(newProps) {
```
sdunster: i'm not going to block this PR if you think it improves it but...
Why do we implement this at all? This should be able to be done in render with low CPU penalty. We should avoid storing state in the component unless absolutely necessary or we actually have a bad perf problem. Leave it for now - I'll have a go at sorting it out later once more critical stuff is done.

sdunster:
I fixed your title

client/src/screens/events/Detail.js
OSPFNeighbour: If we hold it in state then the individual map makers can be notified of a state change if needed.

If we hold it in 'this' or generate it at render time we are pushing new markers every time.

Well thats my understanding and that was done in a PR long ago, unrelated to this specific change directly.

sdunster: I don't think that's how it works. As long as the markers you're rendering are exactly the same (attributes/properties) there should be no need to re-init the map.

PR #455, opened 2018-05-25
Super Awesome Push Token Improvements

OSPFNeighbour:
Updated server and client GQL to return a device when updating device (will make the cache happier)

Changed the way FCM was registering. I think returning a promise of the registration was bad, using the await should be better.

Added some more verbose log messages for tokens.

Added a direct cache update to the UpdateDevice mutation. I was seeing that calling this mutation caused PushManager to run again before the previous mutation had finished, which still had the old token and caused a massive loop. Directly writing the new value into the device prop seems to have fixed this.

Added some control to the push component to stop it from updating its self due it technically updating a prop from within its self it was causing an update etc. Wont update if the new prop device token is the same as the one it just sent to the server.

### DIFF
--- a/client/src/push/fcm.js
+++ b/client/src/push/fcm.js
@@ -56,22 +56,30 @@ class FCMClient {
   }
 
   async register() {
-    const self = this;
-
     // the FCM token might change over time so we register here for updates and push them through
     // to the pushManager
     this.refreshTokenListener = this.FCM.onTokenRefresh((token) => {
-      self.token = token;
-      self.pushManager.updateToken(self.serviceName, token);
+      this.didRegister(token);
     });
 
     // here we return a Promise that fulfills once the token becomes available but we're still
     // responsible for sending it through to the pushManager
-    return this.FCM.getToken().then((token) => {
-      self.token = token;
-      self.pushManager.updateToken(self.serviceName, token);
+    this.FCM.getToken().then((token) => {
+      this.didRegister(token);
+    });
+
+    // block until we get a push token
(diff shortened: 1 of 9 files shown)


client/src/push/fcm.js
```diff
@@ -56,23 +56,31 @@ class FCMClient {
...
+    await new Promise((resolve, reject) => {
+      this.gotToken = resolve;
+      this.failedToGetToken = reject;
     });
   }
 
+  didRegister = (token) => {
+    console.log('Registered for push: ', token);
+    this.pushManager.updateToken(this.serviceName, token);
+    this.gotToken(token);
```
sdunster: you might resolve the promise more than once here, is that a problem?

client/src/push/manager.js
```diff
@@ -45,9 +45,6 @@ class PushManager {
   updateToken(service, token) {
     console.log(`Got token for service ${service}: ${token}`);
     this.tokens[service] = token;
-    if (this.hasAllTokens) {
```
sdunster: you sure we want to remove this?

sdunster:
fix travis

OSPFNeighbour:
Re future updates : FCM will call that back and update the array of tokens, so that should be fine.

sdunster:
updating the array of tokens won't trigger an update though - you also need to call `onTokenUpdate`

OSPFNeighbour:
My research has shown that it does trigger an update, which makes sense because we pass the whole object as a prop to the Token component which reacts to the change in that objects array

PR #475, opened 2018-06-01
Unsub from FCM as per the documentation

OSPFNeighbour:
Fixes "undefined is not a function (evaluating 'this.refreshTokenListener.remove()')" as reported by Bugsnag.

Honestly FCM seems to have no way of turning it off or removing your token so we cant really unregister.

### DIFF
--- a/client/src/push/fcm.js
+++ b/client/src/push/fcm.js
@@ -93,7 +93,7 @@ class FCMClient {
   async deregister() {
     // here we tell firebase we don't care about further updates to the push token
     if (this.refreshTokenListener) {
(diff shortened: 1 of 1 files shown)


client/src/push/fcm.js
```diff
@@ -93,7 +93,7 @@ class FCMClient {
   async deregister() {
     // here we tell firebase we don't care about further updates to the push token
     if (this.refreshTokenListener) {
-      this.refreshTokenListener.remove();
+      this.refreshTokenListener();
```
sdunster: your description says you intend to remove it but the code says otherwise. Why do you call this?

PR #477, opened 2018-06-04
Group Screens for Web

sjmcculloch:
- New component for Tag selection by single or multi
- New message component for displaying standard messages
- View group screen with filters for location, capability, text and my filters
- Edit screen allowing name, location, capabilities and multiple users at once
- Extended server implementation to handle tags/users on group add/edit
- GroupTable, ScheduleTable components
- Centralised styles to AppStyle.js
- New selectors for groups

image

image


web/src/screens/private/Groups/EditGroup.js
```diff
@@ -3,39 +3,328 @@ import PropTypes from 'prop-types';
...
+      this.setInitialState(nextProps);
+    }
+  }
+
+  setInitialState(props) {
+    const group = props.user.organisation.groups.find(
+      g => g.id === parseInt(props.match.params.id, 10),
+    );
+    if (group !== undefined) {
+      let regions = '';
```
OSPFNeighbour: do we have a plan on how to handle the different name we might want to call this? org_structure vs regions vs zone vs units etc etc?

OSPFNeighbour:
That list of users will get pretty crazy with a few thousand in it

server/src/logic.js
```diff
@@ -6,33 +6,27 @@ import { JWT_SECRET } from './config';
...
 // reusable function to check for a user with context
-const getAuthenticatedUser = ctx => ctx.user.then((user) => {
-  // null means we couldn't find the user record
-  if (!user) {
-    // eslint-disable-next-line prefer-promise-reject-errors
-    return Promise.reject('Unauthorized: Invalid user ID');
-  }
-  return user;
-});
+const getAuthenticatedUser = ctx =>
```
sdunster: again with the whitespace changes making it really hard to review.. if you're going to do this stuff I'd really appreciate putting it in a separate commit so I can review it separately.

sdunster: ok I got about halfway through and gave up. Can you comment on the changes?

web/src/screens/private/Events/EditEvent.js
```diff
@@ -246,172 +254,167 @@ class EditEvent extends React.Component {
...
-                      activeStep !== 0 &&
-                      this.state.groupId !== 0 &&
-                      <span> : {this.state.groupName}</span>
-                    }
-                    {index === 1 && activeStep !== 1 && this.state.name !== '' && <span> : {this.state.name}</span>}
-                  </StepLabel>
-                  <StepContent>
-                    <Typography>{getStepContent(index)}</Typography>
-                    {index === 0 && (
+      <div className={classes.root}>
```
sdunster: this is really really long and probably should be broken up into smaller components

web/src/screens/private/Groups/EditGroup.js
sdunster: yeah I thought the whole point of this was that we didn't need to hard code these kinds of things :(
Don't want to block progress at this stage but think about it some more if you can.

web/src/screens/private/Groups/ViewGroups.js
```diff
@@ -3,149 +3,198 @@ import PropTypes from 'prop-types';
...
+              justifyContent: 'space-between',
+              alignItems: 'flex-end',
+            }}
+          >
+            <FormControlLabel
+              control={
+                <Switch
+                  checked={this.state.myGroups}
+                  value="myGroups"
+                  onChange={this.handleSwitchChange('myGroups')}
```
sdunster: i don't think you want to call the function as part of the render function here

web/src/screens/private/Groups/ViewGroups.js
```diff
@@ -3,149 +3,198 @@ import PropTypes from 'prop-types';
...
+                />
+              }
+              label="Only My groups"
+            />
+            <div>
+              <FormControl className={classes.formControlFilter}>
+                <Tag
+                  list={locations}
+                  placeholder="Select Location"
+                  onChange={this.handleTagChange('locationFilter')}
```
sdunster: as above

web/src/screens/private/Groups/ViewGroups.js
```diff
@@ -3,149 +3,198 @@ import PropTypes from 'prop-types';
...
+                  value={this.state.locationFilter}
+                  multi={false}
+                />
+              </FormControl>
+              <FormControl className={classes.formControlFilter}>
+                <Tag
+                  list={capabilities}
+                  placeholder="Select Capability"
+                  onChange={this.handleTagChange('capabilityFilter')}
```
sdunster: same

web/src/selectors/groups.js
```diff
@@ -0,0 +1,32 @@
+export default (
```
sdunster: if we're not going to name this function can you add a comment explaining what it does?

web/src/screens/private/Groups/EditGroup.js
```diff
@@ -3,39 +3,328 @@ import PropTypes from 'prop-types';
...
+              margin="normal"
+              placeholder="Enter Group..."
+            />
+          </FormControl>
+          <FormControl className={classes.formControl}>
+            <Tag
+              list={regions}
+              label="Regions"
+              placeholder="Select..."
+              onChange={this.handleTagChange('regions')}
```
sdunster: dont call function in render

web/src/screens/private/Groups/EditGroup.js
```diff
@@ -3,39 +3,328 @@ import PropTypes from 'prop-types';
...
+              value={this.state.regions}
+              multi
+            />
+          </FormControl>
+          <FormControl className={classes.formControl}>
+            <Tag
+              list={capabilities}
+              label="Capabilities"
+              placeholder="Select..."
+              onChange={this.handleTagChange('capabilities')}
```
sdunster: as above

web/src/screens/private/Groups/EditGroup.js
```diff
@@ -3,39 +3,328 @@ import PropTypes from 'prop-types';
...
+              value={this.state.capabilities}
+              multi
+            />
+          </FormControl>
+          <FormControl className={classes.formControl}>
+            <Tag
+              list={users}
+              label="Users"
+              placeholder="Select..."
+              onChange={this.handleTagChange('users')}
```
sdunster: as above

server/src/logic.js
sjmcculloch: whitespaces? I have some auto-formatting in vscode if that's what is causing the issue

web/src/screens/private/Events/EditEvent.js
sjmcculloch: I agree - I was just matching styles and not really touching this area

web/src/screens/private/Groups/EditGroup.js
sjmcculloch: e.g. I did similar things in SignUpPage.js etc a long time ago 


            <TextField
              required
              fullWidth
              id="password"
              label="Password"
              type="password"
              margin="normal"
              value={this.state.password}
              onChange={this.onPasswordChange}
            />

web/src/screens/private/Groups/EditGroup.js
sjmcculloch: I'm getting a ton of pressure to get this functionality in, I can move it to config in the future but under time pressure at the moment to get these done.

web/src/screens/private/Groups/EditGroup.js
sjmcculloch: nevermind - worked out the issue in the react docs

server/src/logic.js
sdunster: this whole file has changed so much I couldn't be bothered to pick out the actual (non-whitespace) changes. Can you highlight them with inline comments?

web/src/screens/private/Groups/ViewGroups.js
sjmcculloch: For hardcoding - the plan is to use config at an org level for what tags are enabled, I can move this at a later date - trying to balance delivering something and keeping it generic at the moment.

sjmcculloch:
Fixing the function references in next PR.

PR #483, opened 2018-06-05
make refetch not required on home.

OSPFNeighbour:
fixes that exception when you log out.


OSPFNeighbour:
My hunch is its because we have a skip on the apollo query that means it wont run if there is no logged in user. when it doesnt run it looks like the refetch function isnt defined.

PR #486, opened 2018-06-06
Server - Sanity check startTime and endTime to ensure start is before end

### LINKED ISSUE #445: An availability request can be created with the end date being before the start date

OSPFNeighbour:
Tested on server.

Fixes #445

### DIFF
--- a/server/src/creators.js
+++ b/server/src/creators.js
@@ -20,6 +20,9 @@ export const getCreators = (models) => {
       if (typeof startTime === 'undefined' || typeof endTime === 'undefined') {
         return Promise.reject(Error('Must pass start and end times'));
       }
+      if (endTime < startTime) {
+        return Promise.reject(Error('endTime must be greater than startTime'));
+      }
       return Schedule.create({
         name,
         details,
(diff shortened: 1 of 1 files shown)


server/src/creators.js
```diff
@@ -20,6 +20,9 @@ export const getCreators = (models) => {
       if (typeof startTime === 'undefined' || typeof endTime === 'undefined') {
         return Promise.reject(Error('Must pass start and end times'));
       }
+      if (endTime < startTime) {
+        return Promise.reject(Error('endTime must be greater than startTime'));
```
sdunster: you should probably specify:
`endTime must be greater than or equal to startTime`
or change the if statement to use `<=`.

server/src/creators.js
```diff
@@ -122,6 +125,9 @@ export const getCreators = (models) => {
...
```
sdunster: as above

server/src/creators.js
OSPFNeighbour: Err, dont you mean i would have to specify AND change the statement.

> 100 <= 100
true
> 100 < 100
false

thus, endTime < startTime means "endTime must be greater than startTime" an equal endTime and startTime would be true and pass this test, thus my error message is correct assuming we are happy to allow equal times.

Issue #79, opened 2017-11-28
[server] Write health check endpoint


sdunster:
For:
* load balancer instance health checks
* testing of new deployments
* blocking integration tests until the web server is ready

It should:
* return 'not ready' when the web server has started up but the test data hasn't loaded yet
* return 'ready' when the test data has loaded and it is able to query the DB
* return 'unhealthy' if it cannot query the DB or anything else doesn't look good.

We would treat any other kind of error in hitting the endpoint as 'unhealthy'.

Issue #80, opened 2017-11-28
[server] make integration tests wait until the health check passes


sdunster:
Blocked by #79

PR #487, opened 2018-06-06
Server - health check

OSPFNeighbour:
if you hit up /healthcheck it will return something like

```
{"dbConnected":true,"dbDataLoad":true,"uptime":226}
```

I havn't found a way to make sqlite fail because linux is smarter than I am. Sequelize health check method was taken from their official documentation. http://docs.sequelizejs.com/manual/installation/getting-started.html#test-the-connection

#79 Tick

### DIFF
--- a/server/src/main.js
+++ b/server/src/main.js
@@ -20,7 +20,7 @@ const GRAPHQL_PATH = '/graphql';
 
 const port = process.env.PORT ? process.env.PORT : GRAPHQL_PORT;
 
-const { models } = setupDb();
+const { models, db } = setupDb();
 const creators = getCreators(models);
 const { User, Device } = models;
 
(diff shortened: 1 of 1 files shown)


sdunster:
Awesome but to close #79 it also has to fail if the test data is not yet initialised. This way we can delay the integration tests until we know the server is ready.

server/src/main.js
```diff
@@ -81,6 +81,29 @@ app.use(
...
+  let status = 500;
+  const response = {};
+  response.uptime = Math.round(process.uptime());
+  try {
+    db
+      .authenticate()
+      .then(() => {
+        status = 200;
+        response.dbConnector = 'healthy';
+        res.status(status).json(response);
```
sdunster: i think you can probably just omit the status part of this - just let express return what it thinks is sane.

OSPFNeighbour: run the risk of a unhealthy result returning a code 200 if you do that.

OSPFNeighbour: (cause we would want a 500 or something if there is a problem)

OSPFNeighbour:
Technically #79 is only the server side and #80 is the test side.

PR #493, opened 2018-06-08
Client - Change android launch mode

### LINKED ISSUE #492: Loading.js componentDidUpdate
Opening an issue to capture some discussion.

It looks like componentDidUpdate() does not get called when the app is resumed by actioning a notification in Android.

This presents its self as the app will just sit at a loading indicator and never move to the home screen.

I thew a bunch of debug in and it looks like componentDidUpdate() is never called in auth/Loading.js which means we never check if the user is logged in, which means we stay at the <Progress /> and never tell the nav stack to load.

If you throw the auth check into WillMount() it technically fixes the issue but you get a bad warning about updating state on an unmounted component.

To reproduce (On android anyway):

1. Fire off a test push message with a 5 second delay
2. Hide the app
3. Click the notification that pops up in the notification bar

OSPFNeighbour:
set launchMode to Single

Fixes #492

### DIFF
--- a/client/android/app/src/main/AndroidManifest.xml
+++ b/client/android/app/src/main/AndroidManifest.xml
@@ -14,7 +14,8 @@
         android:name=".MainActivity"
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
-        android:windowSoftInputMode="adjustResize">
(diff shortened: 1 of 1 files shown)


OSPFNeighbour:
@sdunster will this need a new shell?

sdunster:
yes it will

Issue #448, opened 2018-05-21
Events: Having selected an Event from the Home Screen doesn't allow the user to go back to the home screen


SSimong:
This is on iPhone.
To Reproduce:

1. open app
2. from home screen, tap on VR at Gerringong falls (one of the events).
3. Event screen opens. Home icon is still selected
4. Tap on Home icon at the bottom. Nothing happens (Left arrow on top left works ok). 
I'd expect the home to allow me to go home.

sdunster:
Agreed this should (and previously did) work as you expected.

PR #497, opened 2018-06-11
tabBarOnPress implemented on TabNavigator to reset Home stack

OSPFNeighbour:
Pressing the Home button will empty the home stack (basically takes you back home)


sdunster:
even with the client upgrades etc this is still necessary?

client/src/navigation/MainNavigator.js
```diff
@@ -43,11 +44,21 @@ const MainTabNavigator = TabNavigator(
...
-      },
+        tabBarOnPress() {
+          const resetAction = NavigationActions.reset({
+            index: 0,
+            key: 'Home',
+            actions: [
+              NavigationActions.navigate({ routeName: 'Root' }),
+            ],
+          });
+          navigation.dispatch(resetAction);
```
sdunster: did you try ditching the reset and just using `navigation.navigate('Root');`?

sdunster: it might help if you explain what you tried before you came to this being the best option

OSPFNeighbour: This was the first thing I tried, and its what 'the internet' documents as the correct way to reset a stack.

There is currently nothing in our code to reset the stack and nothing documented in stacknav that would suggest it should be resetting. When we swap tabs it should just keep the set of tabs in that stack in place. 

#448

OSPFNeighbour: If you just try nav to root you will end up with a weirdly large stack of screens in your stack i believe as it would just add another home screen to the mix.

OSPFNeighbour:
Bump. come up with a better solution or accept my standard, by the book, recommended way of doing it.

sdunster:
rebased

client/src/navigation/MainNavigator.js
```diff
@@ -43,11 +44,21 @@ const MainTabNavigator = TabNavigator(
...
     Home: {
       screen: HomeNavigator,
-      navigationOptions: {
+      navigationOptions: ({ navigation }) => ({
         tabBarLabel: 'Home',
         // eslint-disable-next-line react/prop-types
         tabBarIcon: ({ tintColor }) => <Icon size={34} name="home" color={tintColor} />,
-      },
+        tabBarOnPress() {
+          const resetAction = NavigationActions.reset({
```
sdunster: could you not do this only if the tab is already selected?

sdunster:
Ok I tested it and here is what I observed. I don't think your description made it clear.

On iOS normally (before this diff)
* if you are deep in the home stack hierarchy (you activate home tab and tap on a notification/action which pushes a view onto the stack) then activate another tab, then activate home again the home tab will activate but the nested view stack will be exactly as it was (deep in the hierarchy).
* tapping home button a second time does not reset the nested stack. For all tabs other than home on iOS this would reset the nested stack to root. I believe this a bug and might just be due to the way we have set up the nested root stack (since it works fine for the other tabs).

After the diff:
* tapping home button any time (no matter which tab is currently selected) results in the view stack inside the home tab resetting to the root view (list of notifications/actions).
* this special case only applies to the home tab.
* I believe this is undesirable for iOS, at least. Instead we should fix the hierarchy so that the nested stack can be reset properly.

I added one more comment inline with a suggested change that I think would satisfy all of my concerns and ensure consistency across platforms although I'd still prefer to avoid the need for it on iOS by fixing the underlying hierarchy problem.

tl;dr I think we should continue to stall this PR and look for a better fix. If you think it is absolutely necessary because navigation is completely undesirable on Android and you wish to make it worse for iOS then I won't stop you from merging it.

client/src/navigation/MainNavigator.js
OSPFNeighbour: I played around with detecting that and its possible.

But found that users would be on another tab and want to go home so press home, and it wouldnt take them home, and that was confusing.

PR #507, opened 2018-06-13
Mobile Availability Improvements

sjmcculloch:
- Added pull down to refresh of availability entry screen
- Fixed a bug to filter user ID of timeSegments on the client (will move to graphql query at a later date)

### DIFF
--- a/client/src/screens/schedules/Detail.js
+++ b/client/src/screens/schedules/Detail.js
@@ -46,7 +46,6 @@ class Detail extends Component {
     modalVisible: false,
     modalStartTime: 0,
     modalEndTime: 0,
-    refreshing: false,
   };
 
   componentDidMount() {
@@ -55,7 +54,7 @@ class Detail extends Component {
     });
   }
 
-  onCloseModal = () => {
+  onModalClose = () => {
     this.setState({ modalVisible: false });
   };
 
@@ -69,6 +68,7 @@ class Detail extends Component {
 
   onPressEdit = () => {
     const schedule = this.getSchedule();
+
     // clear existing timeSegments in selected days
     this.state.selectedDays.forEach((day) => {
       const timeSegments = this.getTimeSegments(day);
@@ -203,6 +203,11 @@ class Detail extends Component {
   };
 
   onRefresh = () => {
+    // reset state on refetch
+    this.setState({
+      selectedDays: [],
+      selectionSegments: defaultSegmentState,
+    });
     this.props.refetch();
   }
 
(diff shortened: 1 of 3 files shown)


client/src/selectors/schedules.js
```diff
@@ -27,13 +27,14 @@ export const isSelectorDisabled = (selectionSegments, status, startTime, endTime
...
 
-export const selectSchedules = (schedules, { startTime, endTime }) => {
+export const selectSchedules = (schedules, { userId, startTime, endTime }) => {
   const filteredItems = [];
   schedules.map(schedule =>
     schedule.timeSegments.map((timeSegment) => {
       if (
         (timeSegment.startTime === 0 || timeSegment.startTime >= startTime) &&
-        (timeSegment.endTime === 2147483647 || timeSegment.endTime <= endTime)
+        (timeSegment.endTime === 2147483647 || timeSegment.endTime <= endTime) &&
```
OSPFNeighbour: constants.js distantFuture

client/src/screens/schedules/Detail.js
```diff
@@ -46,6 +46,7 @@ class Detail extends Component {
     modalVisible: false,
     modalStartTime: 0,
     modalEndTime: 0,
+    refreshing: false,
```
OSPFNeighbour: can you use the loading prop from the apollo query?

client/src/screens/schedules/Detail.js
```diff
@@ -326,7 +331,14 @@ class Detail extends Component {
             />
           </Holder>
         </Modal>
-        <ScrollView>
+        <ScrollView
+          refreshControl={
+            <RefreshControl
```
sdunster: TIL you can override this. But we've never needed to do this before, are you sure it's necessary? Or have we always just done it entirely wrong? I think Tim's suggestion is sound though - we should be able to use the apollo loading status to avoid needing state.

sjmcculloch: This is a stop gap until we have proper subscriptions (people submitting availability on web, being able to easy see it on mobile without going back to home screen for refresh). Thought of it as an extra feature. 

I can switch it to use the loading status though.

client/src/screens/schedules/Detail.js
sjmcculloch: I can try

client/src/selectors/schedules.js
sjmcculloch: Yep this is really old code - I will tidy up

client/src/screens/schedules/Detail.js
sjmcculloch: Loading prop is ready only (all props are)

sjmcculloch:
Updates:

- Constant usage
- Can't swap refreshing to loading as props are read only
- RefreshControl will be removed once subscriptions are enabled
- Added some clearing of state when refresh occurs

sdunster:
👍

PR #508, opened 2018-06-13
Server - use default push notification tones

OSPFNeighbour:


OSPFNeighbour:
APN:

```
sound | String | Include this key when you want the system to play a sound. The value of this key is the name of a sound file in your app’s main bundle or in the Library/Sounds folder of your app’s data container. If the sound file cannot be found, or if you specify defaultfor the value, the system plays the default alert sound.For details about providing sound files for notifications; see Preparing Custom Alert Sounds.
```

FireBase:

```
sound | Optional, string | The sound to play when the device receives the notification.Supports "default" or the filename of a sound resource bundled in the app. Sound files must reside in /res/raw/.
```

PR #511, opened 2018-06-13
Server - Map redraw bug

OSPFNeighbour:
was causing the map to reset and rezoom because it was removing the map for a split second.


OSPFNeighbour:
Yeah its both, This causes the token manager to want to redraw for some weird reason (which we settled down with a shouldComponentUpdate), which causes this to draw and you get a mad loop.

PR #512, opened 2018-06-14
Client - remove progress spinner from event details

OSPFNeighbour:
This was not meant to be in the previous PR.

### DIFF
--- a/client/src/screens/events/Detail.js
+++ b/client/src/screens/events/Detail.js
@@ -302,7 +302,6 @@ class Detail extends Component {
     if (!event && !loading) {
       return (
         <Center>
-          <Progress />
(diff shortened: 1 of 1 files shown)


client/src/screens/events/Detail.js
```diff
@@ -302,7 +302,6 @@ class Detail extends Component {
     if (!event && !loading) {
       return (
         <Center>
-          <Progress />
```
sjmcculloch: Approved - but what happens when the query from eventQuery is loading now - you want it to display unable to load event?

OSPFNeighbour: errr no. first time around (when event is undefined) it will show a progress spinner.

Further loading will just be in the background

PR #513, opened 2018-06-14
Client - minor refactor of events details

OSPFNeighbour:
Created two selectors for making map markers
Renamed existing selector MapDelta to mapDelta


client/src/selectors/eventLocationMapMarkers.js
```diff
@@ -0,0 +1,20 @@
+/*
+  Give it an array of event location objects and it will return and array of
+  react-native-maps/map-marker friendly objects (with some extra data mixed in)
+*/
+
+// eslint-disable-next-line import/prefer-default-export
+export const eventLocationMapMarkers = (eventLocations) => {
+  const mapMarkers = [];
+  eventLocations.forEach((r) => {
```
sjmcculloch: Could this eventually be rewritten to something like:

`
const mapMarkers = eventLocations.filter(loc => (loc.locationLatitude !== null && loc.locationLongitude !== null).map(loc => ({ name, locationLatitude, locationLongitude, icon});
`

client/src/selectors/userResponseMapMarkers.js
```diff
@@ -0,0 +1,27 @@
...
+  react-native-maps/map-marker friendly objects (with some extra data mixed in)
+*/
+
+import moment from 'moment';
+
+// eslint-disable-next-line import/prefer-default-export
+export const userResponseMapMarkers = (filter, responses) => {
+  const mapMarkers = [];
+  responses.forEach((r) => {
+    if (r.locationLatitude !== null && r.locationLongitude !== null) {
```
sdunster: please combine these if statements - they can just be `&&`ed

PR #517, opened 2018-06-16
Server Subscriptions (Event, EventResponse, Message)

OSPFNeighbour:
Server side code to allow web socket graphql subscriptions.

```
subscription {
eventResponse(eventId:3) {
  locationTime
  locationLatitude
  locationLongitude
  detail
  status
}
}
```
Change your status on event  number 3 and you get pushed this

```
{
  "eventResponse": {
    "locationTime": 1529130510,
    "locationLatitude": -33.8323275,
    "locationLongitude": 151.0063851,
    "detail": null,
    "status": "unavailable"
  }
}
```

JWT Auth is implemented and seems to work.

Caveats:
- Using a very very basic queue that wont scale in the slightest. We should swap to Redis or something

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -1,6 +1,8 @@
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
 import { Op } from 'sequelize';
+import { withFilter } from 'graphql-subscriptions';
+import { PUBSUBS } from './constants';
 
 import { JWT_SECRET } from './config';
 import { schedulePerms, eventPerms } from './perms';
@@ -25,7 +27,7 @@ const getAuthenticatedDevice = ctx =>
     return device;
   });
 
-export const getHandlers = ({ models, creators: Creators, push }) => {
+export const getHandlers = ({ models, creators: Creators, push, pubsub }) => {
   const { Group, User, Organisation, Schedule, Event, TimeSegment, EventLocation, Tag } = models;
 
   const handlers = {
@@ -258,8 +260,10 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
             });
           });
       },
-      messages(event) {
-        return event.getMessages();
+      messages(schedule) {
+        return schedule.getMessages({
+          order: [['createdAt', 'DESC']],
+        });
       },
     },
 
@@ -416,6 +420,7 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
                 groupId,
               })
               .then(() => {
+                // TODO connect pubsub push for updated event
                 event.getEventlocations().then((existingLocations) => {
                   if (eventLocations === undefined) return;
 
@@ -454,6 +459,27 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
           }),
         );
       },
+      subscribe() {
+        return withFilter(
+          () => pubsub.asyncIterator(PUBSUBS.EVENT.UPDATED),
+          (payload, args) =>
+            // filter is on the OUTBOUND so check there is outbound data,
+            // check its not the users own update, check its for the correct event
+            // return TRUE to send to the subscriber FALSE to not
+            payload && payload.id === args.eventId,
+        );
+      },
+      eventResponseSubscribe() {
+        return withFilter(
+          () => pubsub.asyncIterator(PUBSUBS.EVENTRESPONSE.UPDATED),
+          (payload, args, ctx) => getAuthenticatedUser(ctx).then(user =>
+            // filter is on the OUTBOUND so check there is outbound data,
+            // check its not the users own update, check its for the correct event
+            // return TRUE to send to the subscriber FALSE to not
+            payload && user.id !== payload.userId && payload.eventId === args.eventId,
+          ),
+        );
+      },
       setResponse(args, ctx) {
         const { id } = args.response;
         return eventPerms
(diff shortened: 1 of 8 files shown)


OSPFNeighbour:
Auth is not working correctly

OSPFNeighbour:
I fixed auth :-) client side error in the end

server/src/logic.js
```diff
@@ -692,6 +753,56 @@ export const getHandlers = ({ models, creators: Creators, push }) => {
         return result;
       },
     },
+    subscription: {
+      // Handles logic for incoming sub connections. (User validity only so far)
+      // NOT the actual data return for the subs they are under each respective
+      // object in the main logic
+      message(baseParams, args, ctx) {
+        const newBaseParams = baseParams;
```
sdunster: i feel like you should probably copy this if you're going to modify it.

server/src/main.js
```diff
@@ -81,6 +85,7 @@ app.use(
   '/graphiql',
   graphiqlExpress({
     endpointURL: GRAPHQL_PATH,
+    subscriptionsEndpoint: `ws://localhost:${GRAPHQL_PORT}${SUBSCRIPTIONS_PATH}`,
```
sdunster: I don't think this is going to work in prod

server/src/main.js
```diff
@@ -120,6 +125,55 @@ app.use('/hook', bodyParser.json(), getCallback('ses-hook', creators, models, pu
 
 const graphQLServer = createServer(app);
 
+// eslint-disable-next-line no-unused-vars
+const subscriptionServer = SubscriptionServer.create({
```
sdunster: if this is unused just don't assign it a name..
```
SubscriptionServer.create({
...
```

server/src/main.js
```diff
@@ -120,6 +125,55 @@ app.use('/hook', bodyParser.json(), getCallback('ses-hook', creators, models, pu
...
+  schema,
+  execute,
+  subscribe,
+  onConnect(connectionParams) {
+    let user;
+    let device;
+
+    if (connectionParams.jwt) {
+      jsonwebtoken.verify(connectionParams.jwt, JWT_SECRET,
```
sdunster: this should definitely 100% be abstracted out to a separate module rather than copy paste from our other auth verification methods. Someone will change one and not the other one day and it will break the world.

server/src/schema.js
```diff
@@ -324,9 +325,16 @@ export const Schema = [
     sendTestPush(vars: SendTestPushInput): Boolean  # send a push notification to the requesting device
   }
 
+  type Subscription {
+    message(eventId: Int,groupId: Int): Message
+    eventResponse(eventId: Int!): EventResponse
```
sdunster: shouldn't we just send down updates to this as part of the event? sure it's a little less efficient but would save some extra code. Any time any part of an event changes (including the responses) just send the new event down.

server/src/subscriptions.js
```diff
@@ -0,0 +1,30 @@
+import { PubSub } from 'graphql-subscriptions';
+import { parse } from 'graphql';
+import { getArgumentValues } from 'graphql/execution/values';
+
+export const pubsub = new PubSub();
+
+export function getSubscriptionDetails({ baseParams, schema }) {
```
sdunster: can you please explain why this function exists in a comment

server/src/schema.js
OSPFNeighbour: It's many times less efficient not just a little

server/src/main.js
```diff
@@ -26,64 +31,104 @@ const { User, Device } = models;
...
+    returnUser = User.findById(DEFAULT_USER_ID);
+    returnDevice = Device.findOne({
+      where: {
+        userId: DEFAULT_USER_ID,
+        uuid: DEFAULT_DEVICE_UUID,
+      },
+    });
+  }
+
+  resolve({ user: returnUser, device: returnDevice });
```
sdunster: you're synchronously returning the promise so there's no need for a promise. If you want to return a promise so that you can make this function async in the future without breaking callers then use `return Promise.resolve({ ... });` at the end instead.

```
const postJWTAuth = ({ id, device }) => {
  ...

  return { user: returnUser, device, returnDevice };
}
```

server/src/main.js
```diff
@@ -26,64 +31,104 @@ const { User, Device } = models;
...
 
-
 const app = express();
 
 app.use(cors());
 
-// `context` must be an object and can't be undefined when using connectors
+// Take incoming JTW and resolve user and device and return it. If not passed then use DEFAULT
+// user and device.
+const postJTWAuth = ({ id, device }) => new Promise((resolve) => {
```
sdunster: I like that you're thinking about function names but I think it should be named after what it does rather than where it is called.. perhaps you could name it `resolveUserDevice`?

server/src/main.js
```diff
@@ -26,64 +31,104 @@ const { User, Device } = models;
...
+  execute,
+  subscribe,
+  onConnect(connectionParams) {
+    // theres no standard auth header in WS so we use jwt connection param
+    return jsonwebtoken.verify(connectionParams.jwt || null, JWT_SECRET,
+      (err, decoded) => {
+        if (err) {
+          return Promise.error('Invalid Token');
+        }
+        return postJTWAuth(decoded).then(({ user, device }) => ({ user, device })).catch(() => Promise.error('Unauthorised'));
```
sdunster: please break the `.then` and `.catch` onto multiple lines

server/src/main.js
```diff
@@ -26,64 +31,104 @@ const { User, Device } = models;
...
+      },
+    );
+  },
+  // unpack the subscription request and load it into context
+  onOperation(parsedMessage, baseParams) {
+    const { subscriptionName, args } = getSubscriptionDetails({
+      baseParams,
+      schema,
+    });
+    // if there is a specific function for this sub use it, otherwise use the deafult
```
sdunster: s/deafult/default

server/src/main.js
sdunster: we can try it I guess? does the web UI connect directly to this or does the graphiql backend connect to it? If it's the former it's not going to work.

server/src/schema.js
OSPFNeighbour: went with this in the end just to speed up dev work

PR #518, opened 2018-06-18
Client websocket connector and event Subscription listening

OSPFNeighbour:
Setup for client to connect to WS. Set to lazy so that it will only connect when needed (otherwise it will connect on launch, which is before we have auth)

Minor schema change to server so we get an ID back.

Ive tested most of it locally on my emulators and its working, need to get some tests done with multiple users running at once to see what happens under load.

### DIFF
--- a/client/src/screens/events/Detail.js
+++ b/client/src/screens/events/Detail.js
@@ -4,12 +4,17 @@ import React, { Component } from 'react';
 import { Text, View, Dimensions, StyleSheet, PermissionsAndroid, Platform, Alert, Linking } from 'react-native';
 import MapView, { Marker } from 'react-native-maps';
 import { graphql, compose } from 'react-apollo';
+import update from 'immutability-helper';
+
 import { connect } from 'react-redux';
 import moment from 'moment';
 import _ from 'lodash';
 
+import { wsLink } from '../../app';
+
 import styles from './styles';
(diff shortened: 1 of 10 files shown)


client/src/screens/events/Detail.js
```diff
@@ -99,7 +142,7 @@ class Detail extends Component {
     this.props.navigation.setParams({
       handleThis: this.mapZoomMe,
     });
-    this.refetchTimer = setInterval(this.onRefresh, 5000);
+    // this.refetchTimer = setInterval(this.onRefresh, 5000);
```
sdunster: i'd rather you remove all code relating to this in one go rather than commenting it. Separate commit would be even more amazing but not necessary.

client/src/screens/events/Detail.js
```diff
@@ -73,9 +78,25 @@ class Detail extends Component {
 
   static getDerivedStateFromProps(newProps) {
     // catch incoming props and generate the marker states
-    const { event, loading, auth } = newProps;
+    const { event, loading, auth, subscribeToMore, navigation } = newProps;
     let responseMarkers = {};
     let eventMarkers = {};
+
+    if (newProps.event) {
```
sdunster: please combine if statements:
```
if (newProps.event && !this.subsscription) {
```

PR #519, opened 2018-06-19
Client - Refactor Client GQL Statements

OSPFNeighbour:
Moved heaps of stuff into fragments


OSPFNeighbour:
I clicked every button in the app. That's testing right?

sdunster:
absolutely - just looking for it noted somewhere

PR #521, opened 2018-06-20
Schedule type & web form changes

sjmcculloch:
- Added schedule type to schedules and the "add new request" form
- Restructured "add new request" form and cleaned up method calls
- Added new components (ButtonPanel, FormGroupPanel, SchedulePriority, ScheduleType, DatePicker, Text)

image

### DIFF
--- a/web/src/components/Panels/ButtonPanel.styles.js
+++ b/web/src/components/Panels/ButtonPanel.styles.js
@@ -1,0 +1,8 @@
+export default () => ({
+  buttonPanel: {
+    padding: 20,
+    display: 'flex',
+    alignItems: 'flex-start',
+  },
+});
+
--- a/web/src/screens/private/Schedules/EditSchedule.js
+++ b/web/src/screens/private/Schedules/EditSchedule.js
@@ -9,14 +9,10 @@ import moment from 'moment';
 import { withStyles } from 'material-ui/styles';
 
 import Button from 'material-ui/Button';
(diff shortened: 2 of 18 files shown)


web/src/components/Panels/ButtonPanel.styles.js
```diff
@@ -0,0 +1,8 @@
+export default () => ({
+  buttonPanel: {
+    padding: 20,
+    display: 'flex',
```
OSPFNeighbour: Theres a mix of single quote and double quote strings in your PR. Hadnt noticed it before. the linter on the client forces single quotes on strings.

web/src/screens/private/Schedules/EditSchedule.js
```diff
@@ -130,6 +113,9 @@ class AddSchedule extends React.Component {
       .filter(tag => tag.type === TAG_TYPE_CAPABILITY)
       .map(tag => ({ value: tag.id.toString(), label: tag.name }));
 
+    const groups = user.groups
+      .map(group => ({ value: group.id.toString(), label: group.name }));
```
OSPFNeighbour: toString() will shit a brick if id is every null compared to String() which should not. id should never be null thanks to props validation but something to keep in mind

```
> let value = null
undefined
> value.toString()
TypeError: Cannot read property 'toString' of null
> String(value)
'null'
```

PR #525, opened 2018-06-21
Server - Allow toggle of the DEFAULT user via an ENV variable.

OSPFNeighbour:
if ALLOW_ANON exists the DEFAULT USER will be used, else reject the
connection

### DIFF
--- a/server/src/main.js
+++ b/server/src/main.js
@@ -54,7 +54,6 @@ const postJWTAuth = ({ id, device }) => {
       },
     });
   } else {
-    // If we didnt allow anon access we would reject the promise here
     returnUser = User.findById(DEFAULT_USER_ID);
     returnDevice = Device.findOne({
       where: {
@@ -63,7 +62,6 @@ const postJWTAuth = ({ id, device }) => {
       },
     });
   }
-
   return Promise.resolve({ user: returnUser, device: returnDevice });
 };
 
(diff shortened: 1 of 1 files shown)


server/src/main.js
```diff
@@ -85,8 +83,9 @@ app.use(
       user,
       device,
     },
-  })).catch(() => Promise.reject(Error('Unauthorised'))),
-  ));
+  })).catch(() => false),
```
sdunster: please explain what this changes functionally

server/src/main.js
```diff
@@ -105,11 +104,15 @@ SubscriptionServer.create({
...
     // theres no standard auth header in WS so we use jwt connection param
+    if (!connectionParams.jwt && !process.env.ALLOW_ANON) {
+      return Promise.reject(Error('Unauthorised'));
+    }
     return jsonwebtoken.verify(connectionParams.jwt || null, JWT_SECRET,
       (err, decoded) =>
         postJWTAuth(!err ? { id: decoded.id, device: decoded.device } : { id: null, device: null })
           .then(({ user, device }) => ({ user, device }))
-          .catch(() => Promise.error('Unauthorised')),
+          .catch(() => Promise.reject(Error('Unauthorised')),
```
sdunster: sneaky fix

OSPFNeighbour: Noticed that one did you

server/src/main.js
sdunster: ping

PR #527, opened 2018-06-22
Schedule support for Tags

sjmcculloch:
- Added resolver for schedule -> tags
- Added tags to schedule creator
- Updated test data for schedules

### DIFF
--- a/server/src/schema.js
+++ b/server/src/schema.js
@@ -56,6 +56,7 @@ export const Schema = [
     startTime: Int!
     endTime: Int!
     groupId: Int!
+    tags: [TagInput]
   }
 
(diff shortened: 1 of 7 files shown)


server/src/schema.js
```diff
@@ -276,6 +277,7 @@ export const Schema = [
     startTime: Int!
     endTime: Int!
     timeSegments(userIdFilter: Int): [TimeSegment]!
+    tags: [Tag]! # tags associdated with this schedule
```
OSPFNeighbour: we saying one must have tags?

sjmcculloch: Good pickup - it should be not required

PR #530, opened 2018-06-26
Client - fix race condition on map render

OSPFNeighbour:
this could be empty for a number of reasons


OSPFNeighbour:
Travis didnt see this PR and is kinda fucked. not sure how i force it.

PR #534, opened 2018-06-29
Server - gift-chat fixes

OSPFNeighbour:
Android hates the way this passes state around for onChange. So ive fixed that and will swap back to the main repo once they fix this.

Removed some heavy use of state on the messages handling.

https://github.com/FaridSafi/react-native-gifted-chat/issues/840


OSPFNeighbour:
There is one

PR #538, opened 2018-07-08
race condition on map render

OSPFNeighbour:
Re-created due to travis problems.


sdunster:
Please in the future add more detail in description. When does the race condition occur? Is it consistently reproducible?

PR #539, opened 2018-07-09
Audit logging to DB

OSPFNeighbour:
Module to allow audit logging of "things" to seperate database.

Using a seperate database for flexibility to allow log data to be seperate to the primary data store (cause if that gets fucked over it might be handy to have the audit log survive?)

Anything that was a "Create" I log the returned item
Anything that was an 'Update' I log the requested change

Objects into JSON so that are somewhat standard.

No data relationships to prevent any cascade deleting of data.

Never done this before, I welcome critical comments.

### DIFF
--- a/server/src/logic.js
+++ b/server/src/logic.js
@@ -32,7 +32,7 @@ const areEventNotificationsEnableForUser = (event, user) =>
     { where: { id: user.id } },
   ).then(result => !!result.length);
 
-export const getHandlers = ({ models, creators: Creators, push, pubsub }) => {
+export const getHandlers = ({ logWriter, models, creators: Creators, push, pubsub }) => {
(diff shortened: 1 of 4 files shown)


OSPFNeighbour:
Rebaaaaaaaaaase

server/src/logic.js
```diff
@@ -80,11 +87,18 @@ export const getHandlers = ({ models, creators: Creators, push, pubsub }) => {
       updateUserProfile(_, args, ctx) {
         // TODO support more basic fields
         const { displayName } = args.user;
-        return getAuthenticatedUser(ctx).then(user =>
+        return getAuthenticatedUser(ctx).then((user) => {
           user.update({
```
sdunster: make sure you return this since you changed it to a block from a lambda

OSPFNeighbour: thanks. nice catch

PR #544, opened 2018-07-24
Schedule Time Segment Type for Mobile

sjmcculloch:
Fixes for supplying schedule type -> time segment type on mobile.


client/src/screens/schedules/Detail.js
```diff
@@ -387,16 +388,16 @@ class Detail extends Component {
               <Holder marginTop paddingVertical>
                 <View
                   style={{
-                    flexDirection: 'row',
-                    justifyContent: 'space-between',
-                  }}
+                      flexDirection: 'row',
```
sdunster: revert these whitespace changes? is that your editor?

PR #551, opened 2018-08-01
[Android] Update icons to Callout icons

OSPFNeighbour:
quick and dirty icons.

screenshot 2018-08-01 12 25 11


OSPFNeighbour:
Nah, theres a fancy website that makes the icon bundle for you and it can apply basic image effects.

